PR #100, opened 2015-03-19
correct test error and remove deprecation warning

catawbasam:


stevengj:
Looks like it still has 0.4 compatibility problems.

catawbasam:
Probably should have been more clear.  The aim of this PR is to produce passing tests with no deprecation warnings from Julia 0.3.  It does nothing about problems from Julia 0.4.

Based on my local test and Travis output at `https://travis-ci.org/JuliaLang/JSON.jl/jobs/55005333`, I believe it meets that goal without creating any new problems from Julia 0.4.  Are you seeing something different?

stevengj:
Ah, okay; yes, looks okay.

PR #101, opened 2015-03-21
parseint(s) is deprecated in 0.4

hayd:


src/Parser.jl
```diff
@@ -312,7 +312,7 @@ function parse_number{T<:AbstractString}(ps::ParserState{T})
     if is_float
         float64_isvalid(vs, ps.tmp64) ? (return ps.tmp64[1]) : error("Invalid floating point number", ps)
     else
-        return parseint(vs)
+        return @compat parse(Int, vs)
```
Keno: Base.parse? Does compat handle that?

hayd: I think so: https://github.com/JuliaLang/Compat.jl/pull/53, but this is waiting a release of Compat.

stevengj:
Whoops, I didn't notice this PR and I just pushed a patch to JSON myself (20badf3d11592a31d13deca5a6595cae13e06918)

PR #102, opened 2015-03-24
Fix a bunch of deprecation warnings on v0.4

mweastwood:
This supersedes #101 and should get the Travis builds passing again.


quinnj:
Does this fix the clobbering of Base `parse` method too?

mweastwood:
The `JSON.parse` method is not renamed, but I explicitly qualify `Base.parse` when it's needed.

quinnj:
Cool, looks good.

stevengj:
Whoops, sorry, I just clobbered this patch with 20badf3d11592a31d13deca5a6595cae13e06918 which fixed the `Base.parse` issue ... can you rebase and then we can merge this?

stevengj:
Rebased in #105

PR #104, opened 2015-03-25
Fix #103

### LINKED ISSUE #103: Weird compiler error on latest master?
I'm getting a weird error on latest master; anyone else seeing this or did I mess something up?
```julia
julia> JSON.parse("{1: \"hey\"}")
ERROR: TypeError: instantiate_type: expected TypeConstructor, got Function
 in parse at /Users/jacobquinn/.julia/v0.4/JSON/src/Parser.jl:324
```
I've traced it back to [this line](https://github.com/JuliaLang/JSON.jl/blob/c9cd83c03a35ed29968a7117d7e1e89f22a7ef5d/src/Parser.jl#L230) where I can run the body of the function just fine, but when I try to compile/run the actual method, the error above shows up. As a note, I can also view the output of `@code_lowered` and `@code_typed`, but the error shows up when trying to view `@code_llvm`. See output in [this gist](https://gist.github.com/quinnj/cf1e4e971d549dde8c7e)

@JeffBezanson

quinnj:
Just throw an error when DataStructures isn't present and user tries ordered parsing.

### DIFF
--- a/src/Parser.jl
+++ b/src/Parser.jl
@@ -5,18 +5,10 @@ using Compat
 #Define ordered dictionary from DataStructures if present
 _HAVE_DATASTRUCTURES = try
     using DataStructures
+    import DataStructures.OrderedDict
     true
 catch
     false
-end
-
-if _HAVE_DATASTRUCTURES
-    import DataStructures.OrderedDict
-else
-    function OrderedDict(key_types, types)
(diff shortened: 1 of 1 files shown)


quinnj:
The travis failure looks unrelated (JuliaParser.jl error). Any thoughts? If not, I'll merge in a little while.

kmsquire:
LGTM

quinnj:
I guess technically this is breaking, since without DataStructures, people's code would have still run, but it's also an easy fix and they would have been seeing the warning anyway.

PR #105, opened 2015-03-26
rebased deprecation fixes

stevengj:
Since I broke #102, I felt like I should do the rebase, so here it is.


mweastwood:
Thanks!

PR #113, opened 2015-07-11
Adding JSON macro strings

elcritch:
Basic implementation of JSON macro strings in addition to allowing parsing of non-standard JSON using single quotes `'` around string values. Only single or double quote chars can be used in a single JSON string. 

This is based on the discussion towards the end of [Julia Issue #6739](https://github.com/JuliaLang/julia/issues/6739). 

I've added tests and updated the documentation as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,24 @@ JSON.json(j)
 #  "{\"an_array\":[\"string\",9],\"a_number\":5.0}"
 ```
 
+### Macro Strings
+
+`JSON` and `J` can be used to embed JSON data directly into source code:
+
+```julia
+basic_dict = JSON"""
+{
+  "a_number" : 5.0,
+  "an_array" : ["string", 9]
+}
+"""
+```
+
+
+```julia
+basic_dict = J"{'a_number' : 5.0, 'an_array' : ['string', 9]}"
+```
+
 ## Documentation
 
 ```julia
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -2,7 +2,7 @@ module JSON
 
 using Compat
 
-export json, @J_str, @JSON_str, @JSON_ORDERED_str # returns a compact (or indented) JSON representation as a String
+export json, @J_str, @JSON_mstr, @JSON_ORDERED_mstr # returns a compact (or indented) JSON representation as a String
 
 include("Parser.jl")
 
@@ -283,8 +283,8 @@ function parsefile(filename::AbstractString; ordered::Bool=false, use_mmap=true)
 end
 
 # Macros
-macro JSON_str(arg::AbstractString) parse(arg) end
-macro JSON_ORDERED_str(arg::AbstractString) parse(arg, ordered=true) end
-macro J_str(arg::AbstractString) parse(arg, quote_char='\'') end # convenience intented for short single quoted json data
+macro JSON_mstr(arg::AbstractString) parse(arg) end
+macro JSON_ORDERED_mstr(arg::AbstractString) parse(arg, ordered=true) end
+macro J_str(arg::AbstractString) parse(arg, single_quote=true) end # convenience intented for short single quoted json data
 
 end # module
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -2,7 +2,7 @@ module JSON
 
 using Compat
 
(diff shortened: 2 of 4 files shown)


elcritch:
@ScottPJones , do you know how to get this to run / test on the 0.3 release branch? The 0.4 branch has lot's of deprecation warnings and other issues so I didn't get it tested/working there quite yet.

README.md
```diff
@@ -28,6 +28,26 @@ JSON.json(j)
 #  "{\"an_array\":[\"string\",9],\"a_number\":5.0}"
 ```
 
+### Macro Strings
+
+`JSON` and `J` can be used to embed JSON data directly into source code:
```
IainNZ: wouldn't just `json"""` be more Julian? Either way, :-1: to two ways to do same thing

IainNZ:
So, what is the reason you'd want to have a JSON literal in Julia code?

elcritch:
@IainNZ , basically to allow interoperability with common JavaScript / Python / Ruby code. It's really useful for embedding example configuration examples in Jupyter Notebooks or communicating with Jupyter notebooks.

elcritch:
`json""` and `json""" ... """` would be fine with me. I'll update the code. @kmsquire thanks for the tip re: the postfix modifier. I thought there might be something like that but couldn't figure it out. There is an option in the parser for using `'`. Currently it's either/or as in the parser will only look for double or single quotes.

ScottPJones:
I don't see the need for single quotes, because with triple quotes `"""` you get what you want, double quotes `"` without having to do `\"`

elcritch:
That's a good point. I was trying to go for succinctness originally using `J"{'a':1}"` notation but the `json"..."` is cleaner. Leaving the single quote option but not documenting it would allow flexibility for those who might need it.

README.md
hayd: The documentation here needs to be updated to reflect the code changes.

tshort:
+1 for this feature. See https://github.com/JuliaLang/atom-language-julia/pull/52 for a PR for support in Atom. If we also add support in https://github.com/JuliaLang/Julia.tmbundle (looks easy), we'll get Sublime and GitHub markup covered.

hayd:
RE `json"{'a':1}"`. That's not valid json and should raise, json strings must be `"`-quoted.

It may be useful to have a function that parses python printed dicts but they're not json.

tshort:
I agree, @hayd. The triple-quoted version `json""" ... """` seems most useful to me, anyway.

TotalVerb:
As I mentioned on https://github.com/JuliaLang/JSON.jl/pull/130#issuecomment-225337554, it's kind of pointless that have a string macro that's just a synonym for JSON.parse. It would be nice to have interpolation with this syntax, which can't be done using JSON.parse.

elcritch:
Triple quoted macro is useful for test cases and quick conversion of broken code which uses the deprecated literal dictionary syntax `foo = { "bar": 1 }`. For most any other code, the difference between `JSON.parse(""" ... """)` and `json""" ... """` would be trivial as several mention, but in this case it'd help make up for the loss of built in literal dictionary syntax. Converting existing code is much easier if it's surrounded by just `"""` and not also wrapped in a function call and the trailing `)`. An example situation one would run into this would be the [problem StaffJoy](https://blog.staffjoy.com/retro-on-the-julia-programming-language-7655121ea341#.ytczamafh) folks had where you suddenly have to convert thousands of lines of code to a different syntax, of course string interpolation would be most helpful in that case.

TotalVerb:
@elcritch I don't understand why

```
json"""{
...
}"""
```

is much more painful to write than

```
JSON.parse("""{
...
}""")
```; the difference is only 8 additional characters. In any event, if someone wants this, they can just write

```
macro json_str(ex) Meta.quot(JSON.parse(ex)) end
```

I think, personally, that this is not worth adding until/unless string interpolation is supported.

**Comment by TotalVerb** [comment]
This doesn't actually address the loss of literal dict syntax, because you can't do something like

```
bar = 1
json"""
{
    "foo": bar
}"""
```

and as such `json"x"` is just a different way of spelling `JSON.parse("x")` right now. I think there are better places/packages for dealing with a nicer literal dict syntax.

PR #121, opened 2015-08-30
Add precompile directive

timholy:
Passed `Pkg.test` on 0.4. Needed to resolve dependencies during precompilation.

### DIFF
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -1,3 +1,5 @@
+VERSION >= v"0.4.0-dev+6521" && __precompile__()
+
 module JSON
 
 using Compat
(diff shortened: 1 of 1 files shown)


src/JSON.jl
```diff
@@ -1,3 +1,5 @@
+VERSION >= v"0.4.0-dev+6521" && __precompile__()
```
IainNZ: Not critical, but I've been doing `isdefined(Base, :__precompile__) && __precompile__()`, looks a little less messy

timholy: Not a bad suggestion.

But, I just submitted a whole slew of these, and I'm not gonna redo them without stronger reason than that :smile:.

IainNZ: Hah, yeah, fair enough :D

PR #126, opened 2015-09-25
Fix loading of precompiled JSON when DataStructures is not present

### LINKED ISSUE #123: tag new release?
The main difference from the last tagged version is just the `__precompile__` directive.  It doesn't make a huge difference in the JSON load time at the moment (0.38s vs. 0.58s on my machine), but I'd like to turn on precompilation for IJulia soon (once ZMQ is fixed), and it is good practice to make sure all of the imported modules have `__precompile__` first.

tkelman:
closes #123

first commit adds a test to demonstrate failure case on travis, then second commit moves the conditional dependency on DataStructures into an `__init__` function.

### DIFF
--- a/src/Parser.jl
+++ b/src/Parser.jl
@@ -3,12 +3,13 @@ module Parser #JSON
 using Compat
 
 #Define ordered dictionary from DataStructures if present
-_HAVE_DATASTRUCTURES = try
-    using DataStructures
-    import DataStructures.OrderedDict
-    true
-catch
-    false
+function __init__()
+    global _HAVE_DATASTRUCTURES = try
+        @eval import DataStructures.OrderedDict
+        true
+    catch
+        false
+    end
 end
 
 const TYPES = Any # Union{Dict, Array, AbstractString, Number, Bool, Void} # Types it may encounter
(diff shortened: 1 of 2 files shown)


src/Parser.jl
```diff
@@ -3,12 +3,13 @@ module Parser #JSON
...
 
 #Define ordered dictionary from DataStructures if present
-_HAVE_DATASTRUCTURES = try
-    using DataStructures
-    import DataStructures.OrderedDict
-    true
-catch
-    false
+function __init__()
+    global _HAVE_DATASTRUCTURES = try
```
stevengj: `global const`?

yuyichao: I think Jameson hates it https://github.com/JuliaLang/julia/issues/12010

tkelman: I don't see any harm in allowing this to be modified at runtime if someone really wants to. Would there be a performance impact?

stevengj: Well, globals are usually slower than constant globals because of the type instability...

tkelman: It's only ever used as a boolean afaict - https://github.com/JuliaLang/JSON.jl/blob/0716a1ccc00e63461e03851387390fdcd43d0c35/src/Parser.jl#L322

timholy: You could annotate all usages, e.g., `_HAVE_DATASTRUCTURES::Bool`.

stevengj: Even annotating them is not as efficient as a `const`.  You could make it a `const _HAVE_DATASTRUCTURES = [false]`.

tkelman: I'm going to merge this then, fix what's broken. Adjusting constness can be done separately if it makes a difference.

PR #128, opened 2015-10-05
detect DataStructures.OrderedDict in Pkg.build, not __init__

stevengj:
This avoids the precompilation problem with using `@eval import DataStructures` in `__init__` (noted in fe9de2d226619ea52510479bc30b9acaad77628d) by moving this test from `__init__` to `build` time.  cc @tkelman


tkelman:
Don't think this fixes the error since you could build JSON with datastructures support, remove datastructures, then `using JSON` would fail. Maybe the base stale cache file detection should be modified to not fail fatally when a precompiled dependency is missing, just mark as stale in that case.

stevengj:
Alternatively, we could just `REQUIRE` DataStructures and be done with it.

tkelman:
That would introduce another point of failure for JSON, when many people can get away without the ordered functionality. Perhapa other way around, have DataStructures require JSON so anyone who needs ordered JSON can get it via DataStructures

tkelman:
Or a tiny separate OrderedJSON.jl package that depends on both

stevengj:
Or we could just modify `parse_object` to let you pass in your own `dict` type:
```jl
function parse_object{T<:AbstractString,D<:Associative}(ps::ParserState{T}, dict::Type{D}=Dict)
    parse_object(ps, ordered, dict{KEY_TYPES,TYPES}())
end
```
then you could pass in `OrderedDict` or whatever you want to get any desired order.

tkelman:
I like that idea better.

stevengj:
Of course, that is a breaking change.  But of the 34 packages that currently `REQUIRE` JSON, apparently none of them use the `ordered` keyword option.

stevengj:
Okay, closing this in favor of #129.

PR #129, opened 2015-10-05
change ordered keyword to dicttype keyword

stevengj:
As discussed in #128, this provides more flexibility, and avoids the pseudo-dependency on DataStructures that has been causing precompilation problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,15 +52,18 @@ json(a::Any)
 Returns a compact JSON representation as an `AbstractString`.
 
 ```julia
-JSON.parse(s::AbstractString; ordered=false)
-JSON.parse(io::IO; ordered=false)
-JSON.parsefile(filename::AbstractString; ordered=false, use_mmap=true)
+JSON.parse(s::AbstractString; dicttype=Dict)
+JSON.parse(io::IO; dicttype=Dict)
+JSON.parsefile(filename::AbstractString; dicttype=Dict, use_mmap=true)
 ```
 
 Parses a JSON `AbstractString` or IO stream into a nested Array or Dict.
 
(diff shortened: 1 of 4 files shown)


tkelman:
lgtm. gotta love when adding generality fixes more problems than it adds. worth changing minor pkg version for this, even if apparently unused in public code?

README.md
```diff
@@ -52,15 +52,18 @@ json(a::Any)
 Returns a compact JSON representation as an `AbstractString`.
 
 ```julia
-JSON.parse(s::AbstractString; ordered=false)
-JSON.parse(io::IO; ordered=false)
-JSON.parsefile(filename::AbstractString; ordered=false, use_mmap=true)
+JSON.parse(s::AbstractString; dicttype=Dict)
+JSON.parse(io::IO, dicttype=Dict)
```
IainNZ: `,` should be `;`?

IainNZ:
So very clean!

README.md
stevengj: right, fixed

kmsquire:
While I like this change, I've seen the ordered keyword used in the wild, so the next tag of JSON.jl is going to break some code. Any thoughts on deprecation or informing the user how to fix her code?

stevengj:
@kmsquire, I couldn't find it in any registered packages.

But one could certainly add an `ordered=false` keyword that throws an informative exception if it is changed to `ordered==true`.

kmsquire:
Thanks, yes, I mentioned that this was in user code. I'll add the exception when I get the chance.

tkelman:
sorry for the hasty tag, should have cc'ed you sooner on this

PR #130, opened 2015-11-28
[WIP] Updated json string literal based on elcritch's work

tshort:
The aim of this PR is to update and simplify @elcritch's PR https://github.com/JuliaLang/JSON.jl/pull/113. 

The single-quote option is removed. There's also no option to pass the dicttype. I tried using a trailing argument for this, but I couldn't find an `eval` option that worked. 

It's marked as WIP because there's some extra junk in there for my failed tries to get dicttype to work. Someone else may have a suggestion. In any case, it needs to be cleaned up.

### DIFF
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -6,7 +6,7 @@ using Compat
 
 import Base.colon
 
-export json # returns a compact (or indented) JSON representation as a string
+export json, @json_str, @json_mstr   # returns a compact (or indented) JSON representation as a string
 
 include("Parser.jl")
 
@@ -288,5 +288,18 @@ function parsefile{T<:Associative}(filename::AbstractString; dicttype::Type{T}=D
         JSON.parse(s; dicttype=dicttype)
     end
 end
+
+# Macros 
+str_helper(arg) = parse(arg)
+# I couldn't figure out a way to append the type.
+# The following doesn't work.
+## str_helper(arg, typ) = :(parse($arg, dicttype=eval(symbol($typ))))
+
+macro json_mstr(args...) 
+    str_helper(args...) 
+end
+macro json_str(args...) 
+    str_helper(args...) 
+end 
 
 end # module
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -266,3 +266,18 @@ end
 
 # Check that printing to the default STDOUT doesn't fail
 JSON.print(["JSON.jl tests pass!"],1)
+
+# Tests for Macro strings
+
+let mstr_test = @compat Dict("a" => 1)
+    @test mstr_test == json"""{"a":1}"""
+    @test json"""{"x": 3}"""OrderedDict == DataStructures.OrderedDict{String,Any}([("x",3)])
+
+    @test json"""
+    {
+      "a_number" : 5.0,
+      "an_array" : ["string", 9]
+    }
+    """ == (@compat Dict("a_number" => 5.0, "an_array" => ["string"; 9]) )
+
+end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -271,7 +271,8 @@ JSON.print(["JSON.jl tests pass!"],1)
 
 let mstr_test = @compat Dict("a" => 1)
     @test mstr_test == json"""{"a":1}"""
-    @test json"""{"x": 3}"""OrderedDict == DataStructures.OrderedDict{String,Any}([("x",3)])
+    # Doesn't work:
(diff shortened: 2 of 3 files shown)


tshort:
Any comments on this? Objections to merging?

The test failures are on Julia master and don't relate to this PR.

elcritch:
I'll be able to take a look at it soon. Hopefully tomorrow. 




On Mon, Dec 14, 2015 at 6:13 PM -0800, "Tom Short" <notifications@github.com> wrote:












Any comments on this? Objections to merging?



The test failures are on Julia master and don't relate to this PR.

src/JSON.jl
```diff
@@ -289,4 +289,17 @@ function parsefile{T<:Associative}(filename::AbstractString; dicttype::Type{T}=D
     end
 end
 
+# Macros 
+str_helper(arg) = parse(arg)
```
TotalVerb: this function can probably be taken out

src/JSON.jl
```diff
@@ -6,7 +6,7 @@ using Compat
 
 import Base.colon
 
-export json # returns a compact (or indented) JSON representation as a string
+export json, @json_str, @json_mstr   # returns a compact (or indented) JSON representation as a string
```
TotalVerb: this comment no longer describes the whole line; it should be moved to docstring anyways

src/JSON.jl
```diff
@@ -289,4 +289,17 @@ function parsefile{T<:Associative}(filename::AbstractString; dicttype::Type{T}=D
     end
 end
 
+# Macros 
+str_helper(arg) = parse(arg)
+# I couldn't figure out a way to append the type.
+# The following doesn't work.
+## str_helper(arg, typ) = :(parse($arg, dicttype=eval(symbol($typ))))
+
+macro json_mstr(args...)
```
TotalVerb: I don't think json_mstr is needed

test/runtests.jl
```diff
@@ -266,3 +266,18 @@ end
 
 # Check that printing to the default STDOUT doesn't fail
 JSON.print(["JSON.jl tests pass!"],1)
+
+# Tests for Macro strings
+
+let mstr_test = @compat Dict("a" => 1)
+    @test mstr_test == json"""{"a":1}"""
+    @test json"""{"x": 3}"""OrderedDict == DataStructures.OrderedDict{String,Any}([("x",3)])
```
TotalVerb: this syntax is strange... I'd prefer to just use `JSON.parse` if a special dict type is needed

TotalVerb:
As is, I'm not convinced this is useful—it doesn't save much typing over JSON.parse.

But one feature that would be nice would be interpolation:

```
const x = "World"
json"""
{
    "Hello": $x
}
"""
```

to produce `Dict("Hello" => "World")`.

This interpolation feature is supported by Cxx.jl, MATLAB.jl, and a similar feature has been experimentally implemented in [Scala](http://mandubian.com/2013/07/04/json-interpolation-pattern-matching/).

test/runtests.jl
```diff
@@ -266,3 +266,19 @@ end
 
 # Check that printing to the default STDOUT doesn't fail
 JSON.print(["JSON.jl tests pass!"],1)
+
+# Tests for Macro strings
+
+let mstr_test = @compat Dict("a" => 1)
+    @test mstr_test == json"""{"a":1}"""
+    # Doesn't work:
```
TotalVerb: i would remove the test instead of commenting it out

elcritch:
@TotalVerb having string interpolation in the string macro would be very useful, IMHO. Unfortunately I couldn't find any documentation on how string interpolation is done by on normal Julia parse/compile. It appears you either get string interpolation _or_ you can do string macros.

StefanKarpinski:
String interpolation can't be done in full generality in a string macro since the parser needs to call itself recursively in order to know when the string literal ends. You can approximate it, however, by looking for `$` and using the `parse` function to find an expression. This will not support interpolation of strings.

TotalVerb:
@StefanKarpinski While what you say is entirely correct, this macro is typically used multiline, so string interpolation with string literals (as long as they are not also multiline) should work as expected. I hope to create a prototype soon.

TotalVerb:
As I mentioned [here](https://github.com/JuliaIO/JSON.jl/pull/113#issuecomment-296870619), this macro doesn't really provide anything beyond `JSON.parse("...")`. A different package would be a better place for a replacement for literal dict syntax.

PR #134, opened 2016-02-10
Added simple inline documentation for exported APIs

mdpradeep:


TotalVerb:
This is useful, thanks, sorry it's been left behind so long. Would you mind rebasing this change, as some files have moved around since last year? Otherwise, I could do it for you.

PR #136, opened 2016-02-22
Support Enums

### LINKED ISSUE #135: enums
~~~julia
julia> @enum Test T1 T2 T3

julia> T1
T1::Test

julia> JSON.json(T1)
"{}"
~~~

Is there some way to make this come out like `"\"T1\""` as if I had done `JSON.json("T1")`?

Do I need to override some IO stuff?

Thanks

Edit: I subsequently asked about this on [julia-users](https://groups.google.com/forum/#!topic/julia-users/SQ7KWj69_Gs), Hope you don't mind.

PS: My data structure is actually more complicated and involves enum properties, so hopefully the solution is general enough that it will work for an enum embedded within a more complex type. For example:

~~~julia
julia> d = Dict("a" => T1)
Dict{ASCIIString,Test} with 1 entry:
  "a" => T1::Test

julia> JSON.json(d)
"{\"a\":{}}"
~~~

EricForgy:
Closes #135

### DIFF
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -88,7 +88,7 @@ function _print(io::IO, state::State, s::AbstractString)
     Base.print(io, '"')
 end
 
-function _print(io::IO, state::State, s::Symbol)
+function _print{S <: Union{Symbol,Enum}}(io::IO, state::State, s::S)
     _print(io, state, string(s))
 end
 


src/JSON.jl
```diff
@@ -88,7 +88,7 @@ function _print(io::IO, state::State, s::AbstractString)
     Base.print(io, '"')
 end
 
-function _print(io::IO, state::State, s::Symbol)
+function _print{S <: Union{Symbol,Enum}}(io::IO, state::State, s::S)
```
tkelman: `@compat(Union{Symbol,Enum})`

would also be good to add a test

yuyichao: Also, this doesn't need the type parameter.

EricForgy: Hi @yuyichao 

Are you suggesting it would be better to just make it

~~~julia
function _print(io::IO, state::State, s)
~~~

? Since submitting this, I now don't think it's a great idea to make a special case for enums so I might just close it. What do you think?

yuyichao: I'm saying that it can be `function _print(io::IO, state::State, s::Union{Symbol,Enum})`. I don't have an opinion on if supporting `Enum` this way is necessary or how/if it should be supported.

EricForgy:
I think a better way to deal with this...

If I have MyEnum <: Enum, then just overwrite:

~~~julia
Base._print((io::IO, state::State, s::MyEnum)
~~~

tkelman:
That's not really recommended, and very likely to change (break) in the future. Ref https://github.com/JuliaLang/JSON.jl/pull/151

TotalVerb:
Apologies for the delay on #151. I have something functional locally; I will revisit this tonight and open up a completed version for code review soon.

PR #138, opened 2016-04-13
Correct Date/DateTime

### LINKED ISSUE #90: Datetime not serializing properly.
Given an array of Date objects (from Datetime.jl), `json(dates)` returns `[{}, {}, {}, ..., {}]`, while `json(map(string, dates))` returns the expected `["YYYY-MM-DD", ...]` array. It seems like `string`/`print` isn't being called in all cases where it should in JSON.jl

randyzwitch:
Fixes #90 and https://github.com/johnmyleswhite/Vega.jl/issues/107

Vega.jl examples to show this works in the wild (i.e. parses `Date/DateTime` to a readable format that a JavaScript library will accept and in proper ordinal order)

download 3
download

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -253,6 +253,14 @@ end
 @test sprint(JSON.print, Float64) == string("\"Float64\"")
 @test_throws ArgumentError sprint(JSON.print, JSON)
 
+# test for issue #90 - Date/DateTime
+if VERSION >= v"0.4.6-pre+18"
+@test json(Date("2016-04-13")) == "\"2016-04-13\""
+@test json([Date("2016-04-13"), Date("2016-04-12")]) == "[\"2016-04-13\",\"2016-04-12\"]"
+@test json(DateTime("2016-04-13T00:00:00")) == "\"2016-04-13T00:00:00\""
+@test json([DateTime("2016-04-13T00:00:00"), DateTime("2016-04-12T00:00:00")]) == "[\"2016-04-13T00:00:00\",\"2016-04-12T00:00:00\"]"
+end
+
 # Test parser failures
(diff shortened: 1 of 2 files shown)


quinnj:
+1

test/runtests.jl
```diff
@@ -253,6 +253,14 @@ end
 @test sprint(JSON.print, Float64) == string("\"Float64\"")
 @test_throws ArgumentError sprint(JSON.print, JSON)
 
+# test for issue #90 - Date/DateTime
+if VERSION >= v"0.4.6-pre+18"
```
randyzwitch: Should've asked this before, now CI is going to fail...

What's the convention around versioning within the code? `Dates` came in at 0.4 (?), but I'm developing on nightly `0.4.6`, so do I put the version like I did right here (`"0.4.6-pre+18"`) or do put `0.4` (since presumably, this code would've worked as soon as the `Dates` type was merged into base)? This PR fails for 0.3 because there is no `Dates` type given the current state of my submission.

yuyichao: Check for 0.4 or `isdefined(Base, :Dates)` or conditionally require/import `Dates` should be good enough.

randyzwitch:
The 0.5 failures are unrelated to my changes, but shows my PR failing overall.

randyzwitch:
Any chance we can merge this in and tag a new release? It passes 0.3 and 0.4.

stevengj:
Yes, looks like the failure is unrelated, the fault of the DataStructures package.

stevengj:
In principle we could be using the Dates.jl package to support this in 0.3, but I'd be inclined to just drop support for Julia 0.3 soon.  Looks good to merge.

Issue #127, opened 2015-09-29
Parsing control characters such as newline should be invalid


dreammaker:
Parsing a newline character in a string should be an error, according to the standard and the parsers in basically all other languages.  Using JSON 0.4.6 I get this:

    julia> JSON.parse("\"\n\"")
    "\n"

This has caused us some problems.  We thought our JSON was valid since the Julia JSON package parsed it, but it hasn't been.

stevengj:
Should be a straightforward change to the `parse_string` function (https://github.com/JuliaLang/JSON.jl/blob/fe9de2d226619ea52510479bc30b9acaad77628d/src/Parser.jl#L142).  What characters exactly does the standard permit?

IainNZ:
http://www.json.org/ suggests "anything but control characters, `"`, and `\`"

StefanKarpinski:
Which suggests that JSON is actually doing this right. It seems unreasonable for the package to refuse any syntax that any JSON parser anywhere happens to refuse.

StefanKarpinski:
Oh, does a literal newline count as a control character? In that case, I guess this should be an error.

ScottPJones:
The ECMA standard is a lot clearer:
All characters may be
placed within the quotation marks except for the characters that must be escaped: quotation mark (U+0022),
reverse solidus (U+005C), and the control characters U+0000 to U+001F

TotalVerb:
This is definitely fixed by #140, but I messed up the autoclose so GitHub did not close it. Good to close? @dreammaker @kmsquire

dreammaker:
Closed by #140.

Issue #122, opened 2015-09-05
Parse error


jiahao:
The JSON data in [this gist](https://gist.github.com/jiahao/11b452b42f125e383dcb), returned from a recent GitHub API query, broke the parser:

```jl
julia> JSON.parsefile("events.json")
ERROR: Missing opening string char
Line: 1
Around: ...ntEvent",actor":{"id":1282691...
                    ^

 in error at error.jl:21
 in _error at /Users/jiahao/.julia/v0.4/JSON/src/Parser.jl:70
 in parse_string at /Users/jiahao/.julia/v0.4/JSON/src/Parser.jl:146
 in parse_object at /Users/jiahao/.julia/v0.4/JSON/src/Parser.jl:117
 in parse_object at /Users/jiahao/.julia/v0.4/JSON/src/Parser.jl:107
 in parse_value at /Users/jiahao/.julia/v0.4/JSON/src/Parser.jl:223
 in parse_array at /Users/jiahao/.julia/v0.4/JSON/src/Parser.jl:85
 in parse_value at /Users/jiahao/.julia/v0.4/JSON/src/Parser.jl:223
 [inlined code] from /Users/jiahao/.julia/v0.4/JSON/src/Parser.jl:322
 in __parse#0__ at no file:0
 in parse at no file
 in anonymous at /Users/jiahao/.julia/v0.4/JSON/src/JSON.jl:286
 in open at ./iostream.jl:114
 [inlined code] from /Users/jiahao/.julia/v0.4/JSON/src/JSON.jl:284
 in __parsefile#2__ at no file:0
 in parsefile at no file
```

I suspect it's getting confused by some of the escaped JSON embedded within some of the GitHub issue comments.

TotalVerb:
Are you sure this JSON is not corrupted? There are a lot of control characters in there. I fetched the json file from the link you provided, and it parsed without error.

TotalVerb:
Although this issue wasn't really "fixed" by this #140, the error message seems to have been improved:

before:
```
ERROR: Missing opening string char
Line: 1
Around: ...ntEvent",actor":{"id":1282691...
                    ^
[ backtrace here ]
```

after:
```
ERROR: ASCII control character in string
Line: 0
Around: ...atars.gitubusercontent.com/u/...
                    ^
[ backtrace here ]
```

So I'm inclined to add this to the issues that #140 will close.

TotalVerb:
[...]

Issue #118, opened 2015-08-18
Remove streaming parsing?


IainNZ:
From what I can guess from the tests, these lines: 
https://github.com/JuliaLang/JSON.jl/blob/3067b990f9d58a2088de8cb571bb14954354732c/src/JSON.jl#L211-L280
are a way to consume JSON objects from a stream. Calling `parse` on an `::IO` consumes and returns on JSON.
However, not only does it add a lot of complexity, it does strange things like just silently ignoring bad input, e.g.
```
julia> import JSON

julia> io = IOBuffer("(fwag]pgafdgdf[4,4]")
IOBuffer(data=UInt8[...], readable=true, writable=false, seekable=true, append=false, size=19, maxsize=Inf, ptr=1, mark=-1)

julia> JSON.parse(io)
2-element Array{Any,1}:
 4
 4
```
My proposal would be to just nuke this functionality, as its undocumented and arguably incorrect in the way it tolerates junk silently.

It looks like it was written by @WestleyArgentum, with edits by @jiahao, so would be interested in your take.

WestleyArgentum:
I think the criticisms are totally fair (especially the part where silently tolerates junk), but this was super useful when sending json messages over sockets.

I'm trying to think of a better pattern for people who need to do this (one that doesn't involve everyone re-implementing their own slow+buggy buffers from which they parse out json)... but nothing jumps to mind. Might have to go look at other streaming apis for inspiration.

In the mean time, I don't think we should nuke this without coming up with an alternative pattern. I know there are at least a couple code bases (that I don't have access to) where this is relied on...

StefanKarpinski:
Parsing JSON from a stream and writing JSON to a stream incrementally do seem useful but I don't understand why that entails skipping and ignoring junk on the stream.

WestleyArgentum:
Perhaps the problems can be addressed with optional parameters. We could error when something unexpected happens by default, and maybe provide a way to turn that off.

The other glaring problem I see is that there's no way to specify a timeout, so it's quite easy to get stuck forever when trying to parse from an incomplete stream. Maybe that's a limitation of the IO interface we're using, not sure.

TotalVerb:
The streaming parser no longer silently tolerates junk. It's also a lot faster than it was before. Good to close?

kmsquire:
Can be closed pending https://github.com/JuliaLang/METADATA.jl/pull/5380.

Thanks for going through these issues, @TotalVerb.

[...]

PR #140, opened 2016-04-30
Optimize and fix issues with the `Parser` module

### LINKED ISSUE #98: Trying to `parsefile` an empty file triggers mmap exception
```julia
shell> touch empty.json

julia> import JSON

julia> JSON.parsefile("empty.json")
ERROR: memory mapping failed: Invalid argument
 in systemerror at /usr/local/Cellar/julia/0.3.6_1/lib/julia/sys.dylib
 in mmap at mmap.jl:35
 in mmap_array at mmap.jl:110
 in mmap_array at mmap.jl:4
 in anonymous at /Users/idunning/.julia/v0.3/JSON/src/JSON.jl:280
 in open at /usr/local/Cellar/julia/0.3.6_1/lib/julia/sys.dylib
 in parsefile at /Users/idunning/.julia/v0.3/JSON/src/JSON.jl:279
```

TotalVerb:
Overview of changes:

 - Force parse_string to return Compat.UTF8String
 - Don't use IOBuffer objects in parse_string
 - Sprinkled @inbounds and @inline liberally
 - Make default Dict key typed Compat.UTF8String instead of Any
 - Fix some unicode compatibility along the way
 - Folded the streaming parser into the regular parser, using generic implementations (optimizations for number and string parsing in-memory)

Impact:

Speed for parsing both strings and non-strings have improved substantially on my computer. By my tests we are now easily within a factor of 2 from Python's json with or without strings using [@samuelcolvin](https://gist.github.com/samuelcolvin/11101522)'s benchmarks. Note that these benchmarks heavily rely on float parsing which is still slow, and are ASCII-only which obscures the biggest performance gain obtained.

Here are the new benchmark results. Note that these benchmarks have been "over-benchmarked" to some degree, since they were used heavily in deciding which optimizations to pursue. However, they should in theory represent the kinds of data one would parse quite well. Python's builtin json is somewhat unfairly penalized on these benchmarks, because it is pitifully slow on unicode data. ujson on the other hand is quite good at it. However, ujson "cheats" by default on the numeric benchmarks by not parsing floats to best precision; I included the "non-cheating" version too for completeness.

```
Julia 0.5:
 [Bench] canada 0.057967427 seconds
 [Bench] citm_catalog 0.027227154 seconds
 [Bench] citylots 6.882568502 seconds
 [Bench] twitter 0.011117001 seconds
 [Bench] Total (G.M.) 0.10482892077727128 seconds

Python 2 json:
python json canada time: 0.050s
python json citm_catalog time: 0.032s
python json citylots time: 8.600s
python json twitter time: 0.552s
GM 0.295 (2.81 times Julia)

Python 2 ujson:
python ujson canada time: 0.026s
python ujson citm_catalog time: 0.014s
python ujson citylots time: 5.535s
python ujson twitter time: 0.005s
GM 0.056 (0.534 times Julia)

Python 2 ujson (no cheating):
python ujson canada time: 0.045s
python ujson citm_catalog time: 0.014s
python ujson citylots time: 6.687s
python ujson twitter time: 0.005s
GM 0.068 (0.649 times Julia)

Old Julia 0.5
 [Bench] canada 0.100768632 seconds
 [Bench] citm_catalog 0.089031051 seconds
 [Bench] citylots 16.282590906 seconds
 [Bench] twitter 0.043230788 seconds
 [Bench] Total (G.M.) 0.2819005229832309 seconds (2.69 times new Julia)
```

This change involves a rewrite of object, number, and array parsing code, which also makes things more correct. It removes all of the allowed failures for JSON checker.

Breakage:

The streaming parsers no longer keep track of what has been parsed, and so the error messages are now effectively worthless.

Closes #98, #118, #122, #127.


TotalVerb:
A lot of the remaining time for the no-strings test is due to the type instability of `parse_number`. I am wondering whether it is worth it to simply return `Float64` always from this method. JSON itself does not distinguish between integers and decimals, and JavaScript uses `Float64` for everything.

I will try out some speculative typing for arrays, which might still give a small performance boost.

kmsquire:
@TotalVerb, thanks for doing this!  I'll note that this package sometimes doesn't get much attention, so if you find this PR languishing, please bump it once or twice (or more).

nalimilan:
> A lot of the remaining time for the no-strings test is due to the type instability of parse_number. I am wondering whether it is worth it to simply return Float64 always from this method. JSON itself does not distinguish between integers and decimals, and JavaScript uses Float64 for everything.

Makes sense. Type instability is generally a mistake in Julia anyways.

TotalVerb:
This will need an update to support the latest master's String changes. I plan to work on that later tonight.

TotalVerb:
A major issue now is that `nextind` is very slow for the unified `String` type. Somehow the switch from `ASCIIString` to `String` led to over 50% performance regression on this branch. I have a few ideas for regaining some of the lost performance by avoiding `nextind` where unnecessary. (e.g. making `String` from `String` can be done by byte-copying instead of character-copying, exploiting the fact that ASCII bytes only occur in UTF8 if they are actually encoded ASCII codepoints.)

Master does not have nearly as hard a time dealing with the change to `String` (though it too suffers some). But in large part, that is because master uses `+` where `nextind` may be required.

which is the same as the old `UTF8String` type. In fact, this issue is so bad that it has almost wiped out all performance gains relative to master (apparently much better performance was only achieved on `ASCIIString`s before).

There are some places where `nextind` could probably be replaced with a bytewise scan. I'll look into that.

stevengj:
cc @StefanKarpinski, who will be interested to hear of the `String` change impacts.

TotalVerb:
Actually, I think I was mistaken about the source of the performance regression. `nextind` probably isn't at fault.

TotalVerb:
Here is a condensed benchmark that suffers from significant regression:

```
function f(s)
    n = 0
    m = start(s)
    e = endof(s)
    while m ≤ e
        n += Int(s[m])
        m = nextind(s, m)
    end
    n
end
```

Pre-stringapalooza, on `UTF8String` this is over 10 times as slow as on `ASCIIString`. `String` seems to have inherited `UTF8String`'s speed.

yuyichao:
It's not surprising that the new `String` inherits `UTF8String`'s speed since that's exactly what it is.

I've also seen similar effect in https://groups.google.com/forum/?fromgroups=#!topic/julia-users/9zUDciJk878 where using `nextind` shows a 2x slow down. I remember @StefanKarpinski was talking about more efficient iteration on string index and I was expecting it to come in one of the future related changes.

TotalVerb:
I think the major issues (excepting compatibility with v0.3) are sorted out now. I'll see if reusing buffers still makes sense, and then I'll make sure all the code additions are actually having substantial impact.

TotalVerb:
Non-test code has grown by about 20 to 30 lines. (Some of the added lines are comments.) However, given that a very good chunk of the tests that failed before now pass, I think the slight increase in LOC is reasonable. The only functions which have seen their complexity increase are `parse_string` and `parse_simple`. I don't know if there's a way to keep `parse_string` as simple as it was before while maintaining correctness and efficiency, at least until `nextind` performance becomes much better. `parse_simple` on the other hand was behaving quite oddly earlier (accepting `tXXe` as true, which for some reason was even tested?!), and I think it is better to stick to the JSON standard at some small complexity cost.

The only downside to this change (as far as I can see) is that `parse_number` has less range on 64-bit machines, since it now always returns a `Float64`. (Before, it could parse 64-bit integers on 64-bit machines... but would also crash on those same integers on 32-bit machines, so it is debatable whether that behaviour was good.) It might be useful, in a later PR, to let the user supply a custom parse_number function. Especially now that functions can be specialized on, this would be similar to how a dict type can be passed in currently.

v0.3 compatibility is back, but the timing is slightly worse than on newer versions. I don't think that is a big deal.

test/json_samples.jl
```diff
@@ -296,8 +296,7 @@ twitter = "{\"results\":[
      \"next_page\":\"?page=2&max_id=1480307926&q=%40twitterapi\",
      \"completed_in\":0.031704,
      \"page\":1,
-     \"query\":\"%40twitterapi\"}
-}"
```
tkelman: what was happening here?

TotalVerb: There was an extra close brace in the original test, so it was invalid JSON. This change makes `parse` more picky about that. I don't think it's a good idea to silently accept extra characters after the end.

kmsquire: As an aside, using triple-quotes (`"""`) would make this file much more readable.  (I suspect it was originally created before triple-quotes were part of the language.)

TotalVerb:
I was wondering why the array and object parsing code seems slower recently. It turns out that a recent refactor (moving the burden of throwing the error to a helper function) has led to the construction of a bunch of temporary strings. This can be fixed by removing the second argument, but unfortunately the tailored error messages (like `Expected a '{' here`) will be gone too. A macro would work too, and preserve the tailored error messages, but that looks ugly for some reason...

test/json_samples.jl
aviks: > I suspect it was originally created before triple-quotes were part of the language

Indeed it was.

tkelman:
could delay constructing the string until the error gets displayed?

src/Parser.jl
```diff
@@ -1,45 +1,105 @@
 module Parser #JSON
 
 using Compat
+import Compat: String
 
 export parse
 
-type ParserState{T<:AbstractString}
-    str::T
+# FIXME: remove this when @static in Compat
```
tkelman: better to submit this to Compat first

TotalVerb: OK, I'll send a PR.

TotalVerb:
Should `tryparse` maybe also go in Compat.jl? Then we could be rid of almost all compatibility code.

.travis.yml
```diff
@@ -11,7 +11,7 @@ notifications:
 sudo: false
 script:
     - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-    - julia -e 'Pkg.clone(pwd()); Pkg.build("JSON"); Pkg.test("JSON"; coverage=true)';
+    - julia -e 'Pkg.checkout("Compat"); Pkg.clone(pwd()); Pkg.build("JSON"); Pkg.test("JSON"; coverage=true)';
```
tkelman: it won't be installed yet, you'd have to do this after cloning JSON from pwd

not a bad idea re: submitting `tryparse` to Compat, not sure why no one has asked for it yet but if it's useful here then it's bound to be useful somewhere else

TotalVerb: Whoops, thanks for the catch. I'll make a PR for `tryparse` shortly.

TotalVerb: It turns out `tryparse` is already in `Compat`! It's just the 3-argument form (with radix as third argument) that's missing, which is currently used in parsing unicode escapes.

TotalVerb:
I think the `Int`/`Float64` change may be more breaking than I had expected. At least [Transit.jl](https://github.com/russolsen/Transit.jl) seems to rely on the old behaviour.

It's probably best to revert the type-stability change for now, until there's actually a way to specify the number parse function. (However, a good change would be to swap `Int64` for `Int`, so that 32-bit builds do not crash.)

kmsquire:
> I think the `Int/Float64` change may be more breaking than I had expected. At least `Transit.jl` and any package expecting Ints to roundtrip, may be relying on that behaviour.

Since it's not in the JSON standard, packages shouldn't really be relying on it.  A deprecation warning would be appropriate.

(To be fair, I argued differently when originally introducing `OrderedDict`'s...)

TotalVerb:
There is no real way to deprecate producing `Int`s as numbers without deprecating a whole lot of valid JSON. I think I'll run the tests for some packages that depend on JSON.jl, and see if too much breaks.

kmsquire:
Not that we have to follow Python, but I wasn't aware that they converted to int.  I'm more ok with us doing so as well then.

TotalVerb:
I can't reproduce the 0.3 Travis failure :confused:.

TotalVerb:
I restored the int-parsing behaviour. Performance has degraded a little bit, as expected. But overall it's not so bad, and for some reason it looks like compilation time is much lower. We also can get a bit of performance gain from using `DataStructures.OrderedDict`, if users are inclined to squeeze out every last drop of performance.

TotalVerb:
Great news! `nextind` is quite costly in many parts of the code, and optimizing that has improved the geometric mean benchmark results on my machine to *0.17*. This is even lower than before parse_number type stability was reverted.

This particular change might be applicable to Base too, as it's an almost 3-fold speed improvement over the current implementation of `nextind`.

stevengj:
In this particular application, however, it's not clear to me why you need `nextind` at all.   Since parsing JSON only involves looking for ASCII characters and matching substrings, you should be able to do nearly everything by operating on the raw `s.data` byte array, rather than working with Unicode chars.

TotalVerb:
@stevengj Yes, you're right. In fact some of the performance gain comes from avoiding nextind in string and number parsing. Unfortunately these improved versions somewhat duplicate the generic versions, which provide a performance and memory advantage when parsing a stream.

I did consider avoiding it everywhere, but that currently makes code somewhat messy (== @compat UInt('x') everywhere), and would require a rethink of the streaming parser.

But to get more performance, that seems to be necessary. I'll see if I can figure something out.

stevengj:
(Can some of the byte-based parsing optimizations be folded back in to Base?  e.g. number parsing of UTF-8 strings seems like it could always avoid `nextind` as long as only ASCII digits are allowed.)

stevengj:
I guess you could define `const _x = @compat UInt('x')` to simplify the code somewhat.

TotalVerb:
Base already has a fast strtod, which we're using here. No idea about the other stuff.

The consts are a good idea. I'll play around until I have a solution.

TotalVerb:
I got a huge performance bonus by not copying the buffer data when creating `String`s. Unfortunately, the way I did this was somewhat sketchy:

```
ps.s = s
ret = Compat.UTF8String(UInt8[])
b = sizehint!(ret.data, len)
parse_string(ps, b)
ret
```

Is there a better way to use a known-good UTF8 buffer to create a `String` without making a copy?

src/hoisted.jl
```diff
@@ -0,0 +1,84 @@
...
+    # memory from the start.
+    s = ps.s
+    len = parse_string(ps, 0)
+
+    # Now read the string itself
+    ps.s = s
+    ret = Compat.UTF8String(UInt8[])
+    b = sizehint!(ret.data, len)
+    parse_string(ps, b)
+    ret
```
stevengj: Why can't you just do:
```jl
b = UInt8[]
parse_string(ps, sizehint!(b, len))
return Compat.UTF8String(b)
```
?

src/hoisted.jl
```diff
@@ -0,0 +1,84 @@
...
+            c = str[s]
+            if c == LATIN_U  # Unicode escape
+                ps.s = s + 1
+                for c in string(read_unicode_escape!(ps)).data
+                    b = addbyte!(b, c)
+                end
+                s = ps.s
+                continue
+            elseif haskey(ESCAPES, c)
+                c = ESCAPES[c]
```
stevengj: could avoid the double hash-table lookup here by `c = get(ESCAPES, c, 0xff)` and then throwing an error on `c == 0xff`

src/hoisted.jl
```diff
@@ -0,0 +1,84 @@
...
+            else _error("Unrecognized escape sequence '\\$(@compat Char(c))'", ps)
+            end
+        elseif c < SPACE
+            ps.s = s
+            _error("Illegal control character", ps)
```
stevengj: "Invalid" is a better term than "illegal" in places like this.   (We aren't breaking the law.)

src/Parser.jl
```diff
@@ -48,267 +155,288 @@ function _count_before{T<:AbstractString}(haystack::T, needle::Char, _end::Int)
...
+@inbounds function parse_array(ps::ParserState, dictT::Type)
+    result = Any[]
+    incr!(ps)  # Skip over opening '['
+    chomp_space!(ps)
+    if byteat(ps) ≠ ARRAY_END  # special case for empty array
+        while true
+            push!(result, parse_value(ps, dictT))
+            chomp_space!(ps)
+            c = byteat(ps)
+            c == ARRAY_END && break
```
stevengj: Since `c` is never used elsewhere, these two lines can be combined into `byteat(ps) == ARRAY_END && break`

src/Parser.jl
```diff
@@ -48,267 +155,288 @@ function _count_before{T<:AbstractString}(haystack::T, needle::Char, _end::Int)
...
+    if byteat(ps) ≠ OBJECT_END  # special case for empty object
+        while true
+            # Read key
+            key_ = parse_value(ps, dictT)
+            chomp_space!(ps)
+            skip!(ps, SEPARATOR)
+            # Read value
+            value_ = parse_value(ps, dictT)
+            chomp_space!(ps)
+            obj[convert(keyT, key_)] = value_
```
stevengj: why is the `convert` call needed here?

src/hoisted.jl
TotalVerb: This is twice as slow because `String` makes a copy, at least on 0.5.

```julia
julia> const data = repeat(collect(0x01:0x7F); outer=100);

julia> @time str = String(data);
  0.000018 seconds (7 allocations: 12.781 KB)

julia> data[1] = 0x7F
0x7f

julia> str.data[1]
0x01
```

src/hoisted.jl
TotalVerb: I don't know if unicode escapes are a big bottleneck, but this sounds reasonable and wouldn't make things more complicated, ok.

src/hoisted.jl
TotalVerb: But in some countries newlines are illegal! Thanks.

src/Parser.jl
TotalVerb: Great, thanks.

src/Parser.jl
TotalVerb: One test requires `Symbol` keys.

```julia
julia> obj = Dict{Symbol, Any}()
Dict{Symbol,Any} with 0 entries

julia> obj["hello"] = "world"
ERROR: ArgumentError: hello is not a valid key for type Symbol
 in setindex!(::Dict{Symbol,Any}, ::String, ::String) at ./dict.jl:708
 in eval(::Module, ::Any) at ./boot.jl:225
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```

src/hoisted.jl
stevengj: Oh, right.   You would have to:
```jl
ccall(:jl_array_to_string, Ref{String}, (Any,), b)
```
and calling an undocumented C API function doesn't seem any better than mutating the `String` internals.

(There should probably be an `unsafe_array_to_string` function for this.)

TotalVerb: Seems `:jl_array_to_string` makes the intent more clear, so I'll switch to that anyway. Agreed on the `unsafe_array_to_string`.

src/hoisted.jl
```diff
@@ -0,0 +1,91 @@
...
+    @static if Compat.UTF8String == Base.String
+        b = sizehint!(UInt8[], len)
+        parse_string(ps, b)
+        ccall(:jl_array_to_string, Ref{String}, (Any,), b)
+    else
+        ret = String(UInt8[])
+        b = sizehint!(ret.data, len)
+        parse_string(ps, b)
+        ret
+    end
```
stevengj: It doesn't seem worthwhile to have these two cases.   Can we just use the second case, with `ret = Compat.UTF8String(UInt8[])`, in all Julia versions?

In the longer run, there should probably be an `unsafe_string` added to Base, and then to Compat.

TotalVerb: The second case produces garbage, which isn't good. I have a revised version that performs even better on strings.

TotalVerb:
@stevengj Thanks for the review! I think that's about it for the optimizations for this PR. Now I want to see if I can address the code bloat and duplication... for one I think the generic parse_number and parse_string can be made to use the specialized ones.

TotalVerb:
Filed an issue for Base julia regarding `unsafe_array_to_string` here: https://github.com/JuliaLang/julia/issues/16713

TotalVerb:
Why is this non-mergeable, according to Appveyor?

TotalVerb:
I think there is something weird going on with CI here. Travis doesn't seem to be running. I will rebase the PR and see if that helps.

TotalVerb:
This PR was apparently already rebased, so the rebase seems to have done nothing. But CI is running again... maybe just a fluke.

TotalVerb:
Getting there... did some more cleanup and squashed commits. I reorganized the directory structure and moved error message constants out of the functions themselves. Also updated the commit message.

kmsquire:
> I don't plan on making any further major changes, but I will of course address any remaining review comments.

If it's ready for review, it would be good to change `[WIP]` to `[RFC]` in the title.

TotalVerb:
Done.

kmsquire:
One general comment: unless we expect the code base to become much more complicated, I don't think it's necessary to have a `src/Parser` directory.

kmsquire:
Also, it's always seemed a bit odd to me that JSON printing is kept in JSON.jl, but the parser stuff was (originally) moved to `parser.jl`.  I think it would be good to also move the printing code out of JSON.jl to a separate file that's just included in JSON.jl.  (Having the main file just include a bunch of other files is a typical structure in many packages.)

src/Parser/specialized.jl
```diff
@@ -0,0 +1,122 @@
...
+        unsafe_copy!(b, 1, ps.utf8data, s + 1, len)
+        ps.s = s + len + 2
+    else
+        parse_string(ps, b)
+    end
+
+    _String(b)
+end
+
+function predict_string(ps::MemoryParserState)
```
kmsquire: Can you add a short comment here about what this function is doing?  It's possible to figure out, but it wasn't obvious to me from the name.

src/Parser/Parser.jl
```diff
@@ -0,0 +1,398 @@
...
+"""
+Parse an integer from the given bytes vector, starting at `from` and ending at
+the byte before `to`. Bytes enclosed should all be ASCII characters.
+"""
+function int_from_bytes(bytes::Vector{UInt8}, from::Int, to::Int)
+    @inbounds isnegative = bytes[from] == MINUS_SIGN ? (from += 1; true) : false
+    num = @compat Int64(0)
+    @inbounds for i in from:to
+        num = (@compat Int64(10)) * num +
+                @compat Int64(bytes[i] - DIGIT_ZERO)
```
kmsquire: Julia uses `Int` as an alias for `Int32` on 32-bit systems and `Int64` on 64-bit systems.  Is there a reason we shouldn't follow that convention here?

(As noted before, the JSON spec doesn't actually define integers as part of the spec, only "numbers" (which are floats).  We could follow that, but some packages seem to depend on parsing ints.)

TotalVerb: This lets us parse every number representable exactly as a IEEE double-precision floating point (that is, all numbers that JavaScript can emit). A lot of JSON in the wild is produced by JavaScript, so I think this is important.

We could automatically promote to `Int64` if needed, but I'm a little hesitant about that.

src/Parser/Parser.jl
```diff
@@ -0,0 +1,398 @@
...
+        return parse_number(ps)
+    elseif byte == OBJECT_BEGIN
+        return parse_object(ps, dictT)
+    elseif byte == ARRAY_BEGIN
+        return parse_array(ps, dictT)
+    end
+
+    @jsconstant "true" true
+    @jsconstant "false" false
+    @jsconstant "null" nothing
```
kmsquire: It's nice to remove repeated code like this, but because this macro magically uses `byte` and `ps`, someone reading the code has to go look at the `@jsconstant` macro to understand what's happening here (which is just above of course).

First, I'm wondering if this really needs to be a macro?  You save the lookup of a few characters, but the function version might be just as fast.  (Or it might not.)

Anyway, the easiest thing to do would be to add a comment above this line, and possibly rename the macro to `@isjsconstant` or `@is_jsconstant`.  Actually passing in `byte` and `ps` (as you would for a function) would be the next best thing.

kmsquire:
Generally, looks quite good--thanks again for doing this!  I think it should be mergeable after you address (or at least answer) the comments above.  Some of that can wait until future PRs.

TotalVerb:
Thanks for the comments! I'll address them soon. As for the directory structure, this seems to have been the pattern that will be adopted by Base (https://github.com/JuliaLang/julia/pull/16850). Of course, Base is much more complicated of a codebase too, so maybe we don't need this additional level of organization. I must be doing too much work with the JVM these days ;).

The main JSON.jl file is a bit awkward because it is almost entirely printing... except for `parsefile`. Maybe this function should be moved to the `Parser` module? There was more stuff before when the streaming parser was here too, but that's gone now.

src/Parser/specialized.jl
TotalVerb: It's hard to make the comment short, since what the function does is a little strange. I added a docstring for this and the following function.

TotalVerb:
All the issues except for the organizational ones should be addressed (or answered) now. I'm inclined to defer all the organizational issues to later, since they don't seem to be related to this PR.

tkelman:
It would be a lot easier to see the actual changes being made in a relative sense if the file moves vs modifications were at least separate commits. Right now github shows the old file as completely deleted and the new files as created from scratch, but I don't think you rewrote every line, did you?

TotalVerb:
Good point. I'll revert the file moves; those can come back if desired in the future.

TotalVerb:
Of course, I think over the course of this PR most lines in Parser.jl were changed, so the diff isn't that much better.

src/Parser.jl
```diff
@@ -48,267 +123,277 @@ function _count_before{T<:AbstractString}(haystack::T, needle::Char, _end::Int)
...
+    c = advance!(ps)
+    if c == LATIN_T
+        skip!(ps, LATIN_R, LATIN_U, LATIN_E)
+        true
+    elseif c == LATIN_F
+        skip!(ps, LATIN_A, LATIN_L, LATIN_S, LATIN_E)
+        false
+    elseif c == LATIN_N
+        skip!(ps, LATIN_U, LATIN_L, LATIN_L)
+        nothing
```
kmsquire: I think it would be good to add a comment to each condition specifying what it's testing for, e.g.

```julia
    if c == LATIN_T  #  "true"
        ...
```
It's decodable right now... but a comment can just be read.

TotalVerb: Agreed, that would be better. (skipped ci for comment only change)

kmsquire:
LGTM.  I think this can be merged--I'll do so tomorrow if there are no additional comments.

Thanks for all of your work here, @TotalVerb!

TotalVerb:
This will probably need a squash and rebase first.

TotalVerb:
Squashed and rebased.

PR #141, opened 2016-05-10
Ignore *.mem files from allocation tracking

TotalVerb:
Julia saves .mem files in the package source directory by default, and this makes versioning slightly more annoying. It would be good to ignore these files.


jiahao:
Sure, why not.

You can also put `*.cov` and `*.mem` in your global gitignore, but having these locally will help too.

Maybe we should prepare something for https://github.com/github/gitignore ?

PR #142, opened 2016-05-10
Simple benchmarks

TotalVerb:
It would be helpful to be able to more precisely measure the impact of particular optimizations, so I've imported some benchmarks from the MIT-licensed [nativejson-benchmark](https://github.com/miloyip/nativejson-benchmark). This is very rough, but it's a starting point.

I also added some compliance tests. We're not 100% compliant, but these tests will help pinpoint possible regressions, and also 100% compliance is something to work towards.


kmsquire:
Haven't looked through thoroughly, but at a glance, 👍 !

TotalVerb:
Since they were there anyway, I added the roundtrips to the tests.

tkelman:
how about downloading a tarball instead of inflating the repo size so dramatically?

TotalVerb:
@tkelman Sure. Will work on that later today.

AppVeyor on 32-bit is apparently failing because of an overflow. That's not really desirable in any case. I think a reasonable patch in the interim would be to use `Int64` always, since that supports all JSON numbers, and fix the numeric parsing later.

TotalVerb:
On second thought, this problem was discovered by a test but not introduced by this PR, so skipping the test seems more reasonable. It will be fixed if number parsing becomes type-stable.

test/json-checker.jl
```diff
@@ -0,0 +1,35 @@
+# Run JSON checker tests, excluding tests fail01 and fail18
+
+const JSON_DATA_DIR = joinpath(Pkg.dir("JSON"), "data")
```
tkelman: don't use Pkg.dir, the package may be installed and used from somewhere else. `dirname(@__FILE__)` is better

TotalVerb:
Thanks for the comments. I've made it download the benchmark files when/if required. I also added a citylots file which is much larger.

TotalVerb:
Last try didn't fix the problem. This try should.

TotalVerb:
Well, that still didn't work. I'll have another look tomorrow.

TotalVerb:
Finally passes Appveyor.

test/runtests.jl
```diff
@@ -19,11 +20,12 @@ validate_c(c) = begin
                 end
 
 validate_e(e) = begin
+                    T = TypeVar(:T, AbstractString)
```
tkelman: odd, why's this necessary?

TotalVerb: I don't think it is. That's an unrelated change that accidentally made it in. I'll take it out.

TotalVerb:
Looks like all tests are passing now. Anything else to handle?

test/json-checker.jl
```diff
@@ -0,0 +1,41 @@
...
+
+for i in [2:17; 19:33]
+    file = "fail$(lpad(i, 2, 0)).json"
+    filepath = joinpath(JSON_DATA_DIR, "jsonchecker", file)
+    try
+        JSON.parsefile(filepath)
+
+        # should not reach here
+        print_with_color(:red, " [Noncompliant] ")
+        println("Expecting $file to fail, but parsed without error")
```
tkelman: should this cause an error?

TotalVerb: Those tests don't work yet though. I would use @test_broken if that were available, but it hasn't landed yet.

tkelman: ah okay. add a #FIXME comment saying to make it an error in the future?

tkelman:
This LGTM aside from the FIXME thing. Will see if anyone else has any opinions, if not ping again if I haven't merged tomorrow.

TotalVerb:
Added two FIXME comments and some better error testing. Also added two extra tests to make the numbers nice (previously we were skipping 1 and 18).

A lot of these will be fixed by the parser optimization I'm working on, so hopefully we'll be able to remove the excess code soon enough.

test/json-checker.jl
```diff
@@ -0,0 +1,62 @@
...
+    # FIXME: This is a convoluted way of allowing certain tests to fail. Once
+    # ALLOWED_CHECK_FAILURES is empty, we can eliminate the extra code.
+
+    if i ∈ ALLOWED_CHECK_FAILURES
+        try
+            JSON.parsefile(filepath)
+
+            print_with_color(:red, " [Fail] ")
+            println("Test $file parsed successfully; should throw ErrorException.")
+            println(e)
```
tkelman: what's `e` ?

test/json-checker.jl
```diff
@@ -0,0 +1,62 @@
...
+        catch e
+            if isa(e, ErrorException)
```
tkelman: Doesn't need to be fixed right now, but I wonder if the exception types thrown for various invalid data or parsing failures should be made more specific.

test/json-checker.jl
TotalVerb: Oops. Nice catch.

TotalVerb:
These 32-bit issues are kind of annoying to test. I hope this squashes it?

tkelman:
It looks like. What OS do you use?

TotalVerb:
I'm on Linux, but I have access to a Windows box. Both are 64-bit.

tkelman:
You should be able to download the binaries for win32 and linux32 and run them on 64 bit systems. Packages can get confused if there are binary dependencies, but that's not an issue here.

nalimilan:
> how about downloading a tarball instead of inflating the repo size so dramatically?

Does this really make a large difference in cloning time? The files are very small, and less than 300 lines in total.

tkelman:
An earlier version of this PR was adding over 60000 lines. It's fine now.

tkelman:
Thank you for the contribution!

PR #143, opened 2016-05-14
Fix Julia v0.5 deprecations

TotalVerb:
There aren't that many, and it seems they're pretty easily fixed.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,2 @@
 julia 0.3
-Compat 0.7.1
+Compat 0.7.16
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -74,7 +74,7 @@ function print_escaped(io::IO, s::AbstractString)
         c <= '\x7f' ? (unescaped[@compat(Int(c))+1]  ? Base.print(io, c) :
                        c == '\\'       ? Base.print(io, "\\\\") :
                        c == '"'        ? Base.print(io, "\\\"") :
-                       8 <= c <= 10    ? Base.print(io, '\\', "btn"[@compat(Int(c))-7]) :
+                       8 ≤ @compat(UInt32(c)) ≤ 10 ? Base.print(io, '\\', "btn"[@compat(Int(c))-7]) :
                        c == '\f'       ? Base.print(io, "\\f") :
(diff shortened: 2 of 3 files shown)


TotalVerb:
Shame on me for not testing on release.

src/JSON.jl
```diff
@@ -292,7 +293,7 @@ end
 function parsefile{T<:Associative}(filename::AbstractString; dicttype::Type{T}=Dict, use_mmap=true)
     sz = filesize(filename)
     open(filename) do io
-        s = use_mmap ? UTF8String(Mmap.mmap(io, Vector{UInt8}, sz)) : readall(io)
+        s = use_mmap ? @compat(String(Mmap.mmap(io, Vector{UInt8}, sz))) : readall(io)
```
yuyichao: I don't think the `@compat` is necessary here?

yuyichao: And I believe you want `Compat.UTF8String` instead...

src/Parser.jl
```diff
@@ -1,6 +1,7 @@
 module Parser #JSON
 
 using Compat
+import Compat: String
```
yuyichao: This is not used.

src/JSON.jl
```diff
@@ -3,6 +3,7 @@ isdefined(Base, :__precompile__) && __precompile__()
 module JSON
 
 using Compat
+import Compat: String
```
yuyichao: This can be removed with the case below fixed.

REQUIRE
```diff
@@ -1,2 +1,2 @@
 julia 0.3
-Compat 0.7.1
+Compat 0.7.16
```
yuyichao: The version bump is actually not necessary but doesn't hurt either....

TotalVerb:
Thanks a ton for the comments. If this passes CI then I'll squash the commits.

REQUIRE
tkelman: it is for `Compat.UTF8String`, isn't it?

tkelman:
LGTM

REQUIRE
yuyichao: Only for suppressing warnings. It will work without.

PR #144, opened 2016-05-15
Modernize AppVeyor

tkelman:
Also run tests on supported release versions of Julia

Use Pkg.test("JSON") which now handles adding DataStructures in test/REQUIRE


stevengj:
Thanks!

PR #145, opened 2016-05-15
miscellaneous Compat updates

StefanKarpinski:


yuyichao:
Please update your local copy before updating. Also please make sure not to break 0.4.

yuyichao:
Dup of https://github.com/JuliaLang/JSON.jl/pull/143, which starts with basically the same incorrect fix.

PR #147, opened 2016-05-30
Make it more obvious that allowed failures are allowed

kmsquire:
Cc: @TotalVerb

### DIFF
--- a/test/json-checker.jl
+++ b/test/json-checker.jl
@@ -15,7 +15,7 @@ for i in 1:33
         try
             JSON.parsefile(filepath)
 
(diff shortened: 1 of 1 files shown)


kmsquire:
Also noticed that JSON.jl hasn't been tagged since last October.  After merging this, I plan to tag a release.

TotalVerb:
:+1:. On a related note, it would be good to have functionality like this in `Base.Test`, since I imagine it would be useful to add tests for bugs, even if those bugs haven't been fixed yet. I think there's an outstanding PR about that.

PR #148, opened 2016-06-09
Rename “colon” to “separator”

TotalVerb:
The use of `colon` had nothing to do with `Base.colon`, so it isn’t appropriate to extend (and hence pun) it.

This is the same change as https://travis-ci.org/TotalVerb/JSON.jl/builds/136328207, which <s>passes Travis on my account</s> <s>is being built right now</s> passes Travis on my account. I put [ci skip] on this one because there's a fairly long Travis backlog. I can remove that if it's inappropriate.

### DIFF
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -4,8 +4,6 @@ module JSON
 
 using Compat
 
-import Base.colon
-
 export json # returns a compact (or indented) JSON representation as a string
 
 include("Parser.jl")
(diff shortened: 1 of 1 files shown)


src/JSON.jl
```diff
@@ -41,8 +39,8 @@ suffix(::State{NOINDENT}) = ""
 prefix(s::State{INDENT})  = s.prefix
 prefix(::State{NOINDENT}) = ""
 
-colon(s::State{INDENT})  = ": "
-colon(::State{NOINDENT}) = ":"
+separator(s::State{INDENT})  = ": "
```
kmsquire: You can remove the `s` here.  This was probably copy and pasted from `prefix` above (possibly by me), which does need `s` for the first definition.

kmsquire:
Other than the one minor comment (which was about something already present in the code), LGTM.

src/JSON.jl
TotalVerb: Nice catch. I pushed a fix and updated the CI link in OP.

PR #149, opened 2016-06-12
Fix @os_only deprecation warnings

### LINKED ISSUE #146: @unix_only, @windows_only now deprecated
This is used in a test somewhere.

TotalVerb:
Having red text be printed every time `Pkg.test("JSON")` is run is getting tiresome. Compat has the new `is_windows()` function, and these macros are not coming back (https://github.com/JuliaLang/julia/pull/16658).

Also bumped the Compat version to what is appropriate. With this and #140, there is no more red output for tests on release or master.

Closes #146

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,2 @@
 julia 0.3
-Compat 0.7.16
+Compat 0.7.20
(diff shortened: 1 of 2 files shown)


kmsquire:
Thanks!

Issue #150, opened 2016-06-19
Better custom serialization API?


TotalVerb:
A lot of packages are overloading `JSON._print`. From what I understand, this method is supposed to be internal (its leading underscore and all) but that has not stopped [packages from using it](https://github.com/search?l=&q=%22JSON._print%22+language%3AJulia&ref=advsearch&type=Code&utf8=%E2%9C%93).

This is not desirable, because:

 - Packages need to reimplement a lot of serialization code, like how to serialize dicts and arrays.
 - Overloads of `JSON._print` can write _anything_ — including stuff that isn't JSON. This isn't good.
 - Packages tend to depend on implementation details, like what the name of `JSON.separator` is.

I think a better serialization API, similar to Python's, would help lots. For instance:

 - Introduce `JSON.lower`, designed to be overloaded, that converts an object to another "primitive" object that `JSON` can serialize (`Dict`, `Array`, etc.).
 - In the generic `JSON._print` fallback, call `JSON.lower` instead of the current dumping behaviour
 - Add a generic `JSON.lower` fallback that does the current dumping behaviour

Then:

 - Deprecate overloading `JSON._print`. I'm not sure if there's actually a way to do this. At least, going through the GitHub search above is a good start.
 - Document which methods are public. This is especially important because nothing is exported, so there is no implicit understanding of which methods should be used, and which are internal. This will help avoid breakages like PlotlyJS in the future.

tkelman:
Would using the existing general MIME infrastructure in base make any sense for this?

TotalVerb:
I don't think that would solve the problem. I do like the idea of typing into the general MIME infrastructure, but compare these two APIs:

```julia
JSON.lower(x::MyType) = Dict(:color => [x.red, x.blue, x.green])
```

vs

```julia
function Base.show(io::IO, ::MIME"application/json", x::MyType)
    write(io, "{\"color\": [", x.red, ",", x.blue, ",", x.green, "]}")
end
```

and the second does not even support the pretty-printing features—you'd need to check `IOContext` for that.

TotalVerb:
This overall seems like a fairly minimal change. I'll implement a proof-of-concept and submit a PR later today.

PR #151, opened 2016-06-19
Add JSON.lower for a better serialization API

TotalVerb:
Implements the idea in #150. The following is a quick recap (redundant with #150) of why this is needed:

Lots of packages currently overload `JSON._print` to define custom JSON serialization. This is really bad, because:

- Packages need to reimplement a lot of serialization code, like how to serialize dicts and arrays.
- Overloads of JSON._print can write anything — including stuff that isn't JSON. This isn't good.
- Packages tend to depend on implementation details, like what the name of JSON.separator is.
- Packages work independently of the existing parser system, so they do not benefit from updates to serialization code here.

This PR implements:

- Introduce JSON.lower, designed to be overloaded, that converts an object to another "primitive" object that JSON can serialize (Dict, Array, etc.).
- In the generic JSON._print fallback, call JSON.lower instead of the current dumping behaviour
- Add a generic JSON.lower fallback that does the current dumping behaviour
- Rename JSON._print to JSON._writejson, and add an additional deprecated fallback to a new JSON._print with no methods defined

To make review easier, changes are separated into separate commits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,5 +73,5 @@ JSON.lower(p::Point2D) = [x.x, x.y]
 ```
 
 Define a custom serialization rule for a particular data type. Must return a
-value that can be directly serialized; see help for more details. Do **not**
-extend `JSON._print`; that is a clunky API and is now deprecated.
+value that can be directly serialized; see help for more details. Note that
+`JSON._print` is deprecated and will eventually been discontinued.
--- a/README.md
+++ b/README.md
@@ -67,3 +67,11 @@ provide a desired ordering.  For example, if you `import DataStructures`
 package](https://github.com/JuliaLang/DataStructures.jl) is
 installed), you can pass `dicttype=DataStructures.OrderedDict` to
 maintain the insertion order of the items in the object.
+
+```julia
+JSON.lower(p::Point2D) = [x.x, x.y]
+```
+
+Define a custom serialization rule for a particular data type. Must return a
+value that can be directly serialized; see help for more details. Do **not**
+extend `JSON._print`; that is a clunky API and is now deprecated.
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -80,23 +80,23 @@ function print_escaped(io::IO, s::AbstractString)
     end
 end
 
-function _print(io::IO, state::State, s::AbstractString)
+function _writejson(io::IO, state::State, s::AbstractString)
     Base.print(io, '"')
     JSON.print_escaped(io, s)
     Base.print(io, '"')
 end
 
 if isdefined(Base, :Dates)
-    function _print(io::IO, state::State, s::Base.Dates.TimeType)
-        _print(io, state, string(s))
+    function _writejson(io::IO, state::State, s::Base.Dates.TimeType)
+        _writejson(io, state, string(s))
     end
 end
 
-function _print(io::IO, state::State, s::Symbol)
-    _print(io, state, string(s))
+function _writejson(io::IO, state::State, s::Symbol)
+    _writejson(io, state, string(s))
 end
 
-@compat function _print(io::IO, state::State, s::Union{Integer, AbstractFloat})
+@compat function _writejson(io::IO, state::State, s::Union{Integer, AbstractFloat})
     if isnan(s) || isinf(s)
         Base.print(io, "null")
     else
@@ -104,11 +104,11 @@ end
     end
 end
 
-@compat function _print(io::IO, state::State, n::Void)
+@compat function _writejson(io::IO, state::State, n::Void)
     Base.print(io, "null")
 end
 
-function _print(io::IO, state::State, a::Associative)
+function _writejson(io::IO, state::State, a::Associative)
     if length(a) == 0
         Base.print(io, "{}")
         return
@@ -118,14 +118,14 @@ function _print(io::IO, state::State, a::Associative)
     for (key, value) in a
         first ? (first = false) : Base.print(io, ",", suffix(state))
         Base.print(io, prefix(state))
-        JSON._print(io, state, string(key))
+        _writejson(io, state, string(key))
         Base.print(io, separator(state))
-        JSON._print(io, state, value)
+        _writejson(io, state, value)
     end
     end_object(io, state, true)
 end
 
-@compat function _print(io::IO, state::State, a::Union{AbstractVector,Tuple})
+@compat function _writejson(io::IO, state::State, a::Union{AbstractVector,Tuple})
     if length(a) == 0
         Base.print(io, "[]")
         return
@@ -133,51 +133,57 @@ end
     start_object(io, state, false)
     Base.print(io, prefix(state))
     i = start(a)
-    !done(a,i) && ((x, i) = next(a, i); JSON._print(io, state, x); )
+    !done(a,i) && ((x, i) = next(a, i); _writejson(io, state, x); )
 
     while !done(a,i)
         (x, i) = next(a, i)
         Base.print(io, ",")
         printsp(io, state)
-        JSON._print(io, state, x)
+        _writejson(io, state, x)
     end
     end_object(io, state, false)
 end
 
-function _print(io::IO, state::State, a)
+function _writejson(io::IO, state::State, a)
+    # FIXME: Remove this fallback when _print removed.
+    if applicable(_print, io, state, a)
+        Base.depwarn(
+            "Overloads to `_print` are deprecated; extend `lower` instead.",
+            :_print)
+    end
     start_object(io, state, true)
     range = @compat fieldnames(a)
     if length(range) > 0
         Base.print(io, prefix(state), "\"", range[1], "\"", separator(state))
-        JSON._print(io, state, getfield(a, range[1]))
+        _writejson(io, state, getfield(a, range[1]))
 
         for name in range[2:end]
             Base.print(io, ",")
             printsp(io, state)
             Base.print(io, "\"", name, "\"", separator(state))
-            JSON._print(io, state, getfield(a, name))
+            _writejson(io, state, getfield(a, name))
         end
     end
     end_object(io, state, true)
 end
 
 if VERSION < v"0.5.0-dev+2396"
-function _print(io::IO, state::State, f::Function)
+function _writejson(io::IO, state::State, f::Function)
     Base.print(io, "\"function at ", f.fptr, "\"")
 end
 end
 
-function _print(io::IO, state::State, d::DataType)
+function _writejson(io::IO, state::State, d::DataType)
     Base.print(io, "\"", d, "\"")
 end
 
-function _print(::IO, ::State, m::Module)
+function _writejson(::IO, ::State, m::Module)
     throw(ArgumentError("cannot serialize Module $m as JSON"))
 end
 
 # Note: Arrays are printed in COLUMN MAJOR format.
 # i.e. json([1 2 3; 4 5 6]) == "[[1,4],[2,5],[3,6]]"
-function _print{T, N}(io::IO, state::State, a::AbstractArray{T, N})
+function _writejson{T, N}(io::IO, state::State, a::AbstractArray{T, N})
     lengthN = size(a, N)
     if lengthN > 0
         start_object(io, state, false)
@@ -187,12 +193,12 @@ function _print{T, N}(io::IO, state::State, a::AbstractArray{T, N})
             newdims = ntuple(i -> 1:size(a, i), N - 1)
         end
         Base.print(io, prefix(state))
-        JSON._print(io, state, Compat.view(a, newdims..., 1))
+        _writejson(io, state, Compat.view(a, newdims..., 1))
 
         for j in 2:lengthN
             Base.print(io, ",")
             printsp(io, state)
-            JSON._print(io, state, Compat.view(a, newdims..., j))
+            _writejson(io, state, Compat.view(a, newdims..., j))
         end
         end_object(io, state, false)
     else
@@ -200,15 +206,18 @@ function _print{T, N}(io::IO, state::State, a::AbstractArray{T, N})
     end
 end
 
+"Deprecated way to overload JSON printing behaviour. Use `lower` instead."
+function _print end
+
 function print(io::IO, a, indent=0)
-    JSON._print(io, State(indent), a)
+    _writejson(io, State(indent), a)
     if indent > 0
         Base.print(io, "\n")
     end
 end
 
 function print(a, indent=0)
-    JSON._print(STDOUT, State(indent), a)
+    _writejson(STDOUT, State(indent), a)
     if indent > 0
         println()
     end
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -200,15 +200,14 @@ end
 function _writejson(io::IO, state::State, a)
     # FIXME: This fallback is harming performance substantially.
     # Remove this fallback when _print removed.
-    #= if applicable(_print, io, state, a)
+    if applicable(_print, io, state, a)
         Base.depwarn(
             "Overloads to `_print` are deprecated; extend `lower` instead.",
             :_print)
         _print(io, state, a)
     else
         _writejson(io, state, lower(a))
-    end =#
-    _writejson(io, state, lower(a))
+    end
 end
 
 # Note: Arrays are printed in COLUMN MAJOR format.
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -7,12 +7,26 @@ using Compat
 export json # returns a compact (or indented) JSON representation as a string
 
 include("Parser.jl")
+include("bytes.jl")
 
 import .Parser.parse
+
+# These are temporary ways to bypass excess memory allocation
+# They can be removed once types define their own serialization behaviour again
+"Internal JSON.jl implementation detail; do not depend on this type."
+immutable AssociativeWrapper{T} <: Associative{Symbol, Any}
+    wrapped::T
+    fns::Array{Symbol, 1}
+end
+AssociativeWrapper(x) = AssociativeWrapper(x, @compat fieldnames(x))
 
 typealias JSONPrimitive @compat(Union{
         Associative, Tuple, AbstractArray, AbstractString, Integer,
         AbstractFloat, Bool, Void})
+
+Base.getindex(w::AssociativeWrapper, s::Symbol) = getfield(w.wrapped, s)
+Base.keys(w::AssociativeWrapper) = w.fns
+Base.length(w::AssociativeWrapper) = length(w.fns)
 
 """
 Return a value of a JSON-encodable primitive type that `x` should be lowered
@@ -29,13 +43,7 @@ Note that the return value need not be *recursively* lowered—this function may
 for instance return an `AbstractArray{Any, 1}` whose elements are not JSON
 primitives.
 """
-function lower(a)
-    obj = @compat Dict{Symbol, Any}()
-    for name in @compat fieldnames(a)
-        obj[name] = getfield(a, name)
-    end
-    obj
-end
+lower(a) = AssociativeWrapper(a)
 lower(a::JSONPrimitive) = a
 
 if isdefined(Base, :Dates)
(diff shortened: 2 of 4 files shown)


quinnj:
Yes please; this looks great

TotalVerb:
We still need a way to deprecate these overloads. Unfortunately, `@deprecate_binding` won't work. I have an idea I'll play around with later.

TotalVerb:
#152 is caused by trying to reimplement array serialization and not catching all the edge cases (such as the zero-column case). I think it's a good idea to lower multi-dimensional arrays to vectors, so that these issues don't occur again in the future.

TotalVerb:
This is no longer WIP. To make review easier, I've separated the commits into logical units:

 - The first commit deprecates extending JSON._print (deprecation warning printed on use)
 - The second commit implements `lower`, as described in OP, as a replacement
 - The third commit replaces existing `_writejson` overloads to `lower` methods
 - The fourth commit documents the `lower` function in the README

I've verified locally that `JSON` tests and `Plotly` (which overloads `JSON._print`) tests pass on all intermediate commits.

TotalVerb:
I'd love for coverage to not drop, but I don't know how to cover deprecated codepaths without invoking deprecation warnings.

EricForgy:
@malmaud , any thought on this PR? If you like the general direction, I'm happy to do what I can to help out with testing etc. Just tight on time so don't want to test something with no chance of being merged :sweat_smile: 

This seems like a fairly big change to a core package that may be incorporated into Base at some point.

EricForgy:
@essenciary, any thoughts?

src/JSON.jl
```diff
@@ -10,6 +10,47 @@ include("Parser.jl")
...
+for instance return an `AbstractArray{Any, 1}` whose elements are not JSON
+primitives.
+"""
+function lower(a)
+    obj = @compat Dict{Symbol, Any}()
+    for name in @compat fieldnames(a)
+        obj[name] = getfield(a, name)
+    end
+    obj
+end
```
kmsquire: This definition seems quite broad as a fallback.  

For example, if I [define a new bitstype](https://github.com/diegozea/BioSeq.jl/blob/master/src/biounits.jl#L1-L4), it will fallback here, which is probably not that useful.  Maybe this can be restricted to `DataTypes`?  (Not sure...)

kmsquire:
One potential downside of `lower` (compared with `_print`) is that it creates a number of intermediate strings.  

In contrast, `_print` accepts an `IO` argument as its first parameter, and sends the conversion directly there, which is more in line with the way `print` and friends work in Base Julia.

Compared to most languages I've worked with, the choice to work directly with `IO` objects instead of constructing and passing around strings seems a little awkward, but it's probably faster... although I don't have any benchmarks to back it up (Cc: @StefanKarpinski, who I believe was the main driver behind the output convention).

@TotalVerb, you've recently sped up the code here in JSON.jl significantly.  Have you done any benchmarking here to verify that the introduction of intermediate strings isn't causing much slowdown?  Obviously, this would mostly happen for large types which need to be lowered (e.g., `Associatives` with a lot of members).

Anyway, consider recoding using an `IO` argument.

src/JSON.jl
TotalVerb: For what it's worth, this is the same fallback as before. I'll look into restricting it.

TotalVerb:
@kmsquire You are very right about the possible performance impact. However,

 - The only package on GitHub whose overload of `JSON._print` will be slowed down is `Plots`. The other three packages are creating temporary strings anyway.
 - The original code in this package itself was frequently creating temporary strings. I suspect (but have not measured) that serializing the other types that are supported will not in general be slower. Example:

```julia
if isdefined(Base, :Dates)
    function _print(io::IO, state::State, s::Base.Dates.TimeType)
        _print(io, state, string(s))
    end
end

function _print(io::IO, state::State, s::Symbol)
    _print(io, state, string(s))
end
```

 - `Associative`s, will actually be serialized directly and do not need to be lowered. There is no performance impact to serializing built-in types.
 - I don't see this as the final solution. This is just a way to stop the bleeding, and stop `JSON._print` overloading from getting worse.

I do have some ideas for how to improve the performance of JSON serialization using IO arguments; however, I don't think that should stop us from deprecating `JSON._print` ASAP. (Note that overloading this is the cause of the only non-test failure that arose from JSON v0.5.2.)

src/JSON.jl
kmsquire: I hadn't realized it was the same as before.  FWIW, I doubt custom bitstypes are that common, so it might not matter in a practical sense.

TotalVerb:
Actually, I'm not fully confident myself. Benchmarks would certainly be useful to either confirm that there is no impact, or bring attention to a real performance problem. I'll prepare a few.

kmsquire:
> I don't see this as the final solution. This is just a way to stop the bleeding, and stop `JSON._print` overloading from getting worse.

Agree that overloading `JSON._print` isn't good, and should be deprecated... but after thinking some more, I'm wondering if it wouldn't just be better to rename `JSON._print` to `JSON.print_json`, and allow people to override it.

EricForgy:
It makes me nervous when you say things like that. As I said, I suspect there are a lot of private repos dependent on JSON (I have lots), so we should really be thoughtful of performance impacts (if any).

Get Outlook for iOS<https://aka.ms/o0ukef>




On Tue, Aug 9, 2016 at 1:20 PM +0800, "Fengyang Wang" <notifications@github.com<mailto:notifications@github.com>> wrote:


@kmsquire<https://github.com/kmsquire> You are very right about the possible performance impact. However,

  *   The only package on GitHub whose overload of JSON._print will be slowed down is Plots. The other three packages are creating temporary strings anyway.
  *   The original code in this package itself was frequently creating temporary strings. I suspect (but have not measured) that serializing the other types that are supported will not in general be slower. Example:

if isdefined(Base, :Dates)
    function _print(io::IO, state::State, s::Base.Dates.TimeType)
        _print(io, state, string(s))
    end
end

function _print(io::IO, state::State, s::Symbol)
    _print(io, state, string(s))
end

  *   Associatives, will actually be serialized directly and do not need to be lowered. There is no performance impact to serializing built-in types.
  *   I don't see this as the final solution. This is just a way to stop the bleeding, and stop JSON._print overloading from getting worse.

I do have some ideas for how to improve the performance of JSON serialization using IO arguments; however, I don't think that should stop us from deprecating JSON._print ASAP. (Note that overloading this is the cause of the only non-test failure that arose from JSON v0.5.2.)

-

TotalVerb:
@kmsquire Unfortunately that doesn't solve the following:

- Overloads of JSON._print can write anything — including stuff that isn't JSON. This isn't good.
- Packages tend to depend on implementation details, like what the name of JSON.separator is.
- Packages work independently of the existing parser system, so they do not benefit from updates to serialization code here.

I do have a long-term solution in mind that will improve on the current state of affairs. I need to work out the kinks, but expect a PR soon-ish.

@EricForgy Don't worry. I am creating benchmarks and will make sure that performance does not degrade.

kmsquire:
@TotalVerb, thanks--that makes sense.

If there aren't any comments soon, I would think it fine to merge this.

TotalVerb:
I will finish the benchmarks tonight and report results. If there are any regressions, I suspect they're not difficult to fix.

TotalVerb:
I created a benchmark suite with 19 tests for both compact printing and pretty printing (PR incoming). Here are the results:

```julia
julia> showall(judge(minimum(results["print"]), minimum(last["print"]); time_tolerance=0.05))
19-element BenchmarkTools.BenchmarkGroup:
  tags: ["serialize"]
  "ascii" => BenchmarkTools.TrialJudgement: 
  time:   +0.00% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "flat-homogenous-array-16" => BenchmarkTools.TrialJudgement: 
  time:   +2.09% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "nested-array-16^2" => BenchmarkTools.TrialJudgement: 
  time:   -4.29% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "unicode" => BenchmarkTools.TrialJudgement: 
  time:   -2.73% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "null" => BenchmarkTools.TrialJudgement: 
  time:   -8.49% => improvement (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "unicode-1024" => BenchmarkTools.TrialJudgement: 
  time:   -3.42% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "flat-dict-128" => BenchmarkTools.TrialJudgement: 
  time:   -8.25% => improvement (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "matrix-16" => BenchmarkTools.TrialJudgement: 
  time:   -3.47% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "custom-tree-8" => BenchmarkTools.TrialJudgement: 
  time:   +198.28% => regression (5.00% tolerance)
  memory: +751.81% => regression (1.00% tolerance)
  "small-dict" => BenchmarkTools.TrialJudgement: 
  time:   +25.97% => regression (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "heterogenous-array" => BenchmarkTools.TrialJudgement: 
  time:   -9.40% => improvement (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "float" => BenchmarkTools.TrialJudgement: 
  time:   -5.99% => improvement (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "bool" => BenchmarkTools.TrialJudgement: 
  time:   -1.66% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "flat-homogenous-array-1024" => BenchmarkTools.TrialJudgement: 
  time:   -6.17% => improvement (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "integer" => BenchmarkTools.TrialJudgement: 
  time:   +0.54% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "date" => BenchmarkTools.TrialJudgement: 
  time:   +59.04% => regression (5.00% tolerance)
  memory: +1.37% => regression (1.00% tolerance)
  "custom-list-128" => BenchmarkTools.TrialJudgement: 
  time:   +167.12% => regression (5.00% tolerance)
  memory: +469.86% => regression (1.00% tolerance)
  "ascii-1024" => BenchmarkTools.TrialJudgement: 
  time:   +1.04% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "nested-array-16^3" => BenchmarkTools.TrialJudgement: 
  time:   +0.87% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
julia> showall(judge(minimum(results["pretty-print"]), minimum(last["pretty-print"]); time_tolerance=0.05))
19-element BenchmarkTools.BenchmarkGroup:
  tags: ["serialize"]
  "ascii" => BenchmarkTools.TrialJudgement: 
  time:   -3.26% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "flat-homogenous-array-16" => BenchmarkTools.TrialJudgement: 
  time:   +1.46% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "nested-array-16^2" => BenchmarkTools.TrialJudgement: 
  time:   -2.62% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "unicode" => BenchmarkTools.TrialJudgement: 
  time:   -4.19% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "null" => BenchmarkTools.TrialJudgement: 
  time:   -3.55% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "unicode-1024" => BenchmarkTools.TrialJudgement: 
  time:   +4.10% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "flat-dict-128" => BenchmarkTools.TrialJudgement: 
  time:   -3.74% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "matrix-16" => BenchmarkTools.TrialJudgement: 
  time:   +0.20% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "custom-tree-8" => BenchmarkTools.TrialJudgement: 
  time:   +93.14% => regression (5.00% tolerance)
  memory: +332.62% => regression (1.00% tolerance)
  "small-dict" => BenchmarkTools.TrialJudgement: 
  time:   +18.99% => regression (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "heterogenous-array" => BenchmarkTools.TrialJudgement: 
  time:   +0.18% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "float" => BenchmarkTools.TrialJudgement: 
  time:   +1.30% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "bool" => BenchmarkTools.TrialJudgement: 
  time:   -14.86% => improvement (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "flat-homogenous-array-1024" => BenchmarkTools.TrialJudgement: 
  time:   -2.45% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "integer" => BenchmarkTools.TrialJudgement: 
  time:   +0.43% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "date" => BenchmarkTools.TrialJudgement: 
  time:   +28.43% => regression (5.00% tolerance)
  memory: +1.37% => regression (1.00% tolerance)
  "custom-list-128" => BenchmarkTools.TrialJudgement: 
  time:   +39.32% => regression (5.00% tolerance)
  memory: +142.96% => regression (1.00% tolerance)
  "ascii-1024" => BenchmarkTools.TrialJudgement: 
  time:   -2.59% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
  "nested-array-16^3" => BenchmarkTools.TrialJudgement: 
  time:   -3.33% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
```

Most of the differences are just noise, but there are definitely real regressions. I don't think they will be too hard to fix, however. Going to add some patches.

kmsquire:
👍

TotalVerb:
I eliminated all the time regressions (memory regressions are a little harder, but those should mostly go away after the next PR I have planned)—however, that required removing the deprecated code path, which alone explains most of the time regressions found above. (Turns out `applicable` is pretty expensive, especially when it's called frequently!) I think the deprecation will have to stay, unfortunately, so that packages are not adversely affected.

I also slightly improved the string serialization, so some timings (on tests with long strings) are better than before. Here are the timings with the deprecation warning excised; I removed ones that are invariant to save space, and I consider the remaining ones (except for the memory regressions) either noise or very small changes. Because of the level of noise, I set the time threshold to 10% for all tests. Overall, things look pretty good.

```
[ print ]
  "flat-dict-128" => BenchmarkTools.TrialJudgement: 
  time:   -11.62% => improvement (10.00% tolerance) [not real; this code wasn't changed]
  "custom-tree-8" => BenchmarkTools.TrialJudgement: 
  time:   +11.34% => regression (10.00% tolerance)  [don't think this is real, exaggerated if it is]
  memory: +245.49% => regression (1.00% tolerance)  [real]
  "small-dict" => BenchmarkTools.TrialJudgement: 
  memory: -9.09% => improvement (1.00% tolerance)  [real, hex escapes allocate less]
  "heterogenous-array" => BenchmarkTools.TrialJudgement: 
  time:   -18.01% => improvement (10.00% tolerance)  [possibly real, but definitely exaggerated]
  "date" => BenchmarkTools.TrialJudgement: 
  time:   +17.89% => regression (10.00% tolerance)  [don't think this is real]
  "custom-list-128" => BenchmarkTools.TrialJudgement: 
  memory: +151.87% => regression (1.00% tolerance)  [real]
  "ascii-1024" => BenchmarkTools.TrialJudgement: 
  time:   -22.65% => improvement (10.00% tolerance)  [real]
[ pretty-print ]
  "ascii" => BenchmarkTools.TrialJudgement: 
  time:   -14.05% => improvement (10.00% tolerance)  [real]
  "matrix-16" => BenchmarkTools.TrialJudgement: 
  time:   +11.54% => regression (10.00% tolerance)  [not real]
  "custom-tree-8" => BenchmarkTools.TrialJudgement: 
  memory: +108.61% => regression (1.00% tolerance) [real]
  "small-dict" => BenchmarkTools.TrialJudgement: 
  memory: -7.19% => improvement (1.00% tolerance)  [real]
  "bool" => BenchmarkTools.TrialJudgement: 
  time:   -17.74% => improvement (10.00% tolerance)  [not real]
  "custom-list-128" => BenchmarkTools.TrialJudgement: 
  memory: +46.21% => regression (1.00% tolerance)  [real]
  "ascii-1024" => BenchmarkTools.TrialJudgement: 
  time:   -15.96% => improvement (10.00% tolerance)  [real]
```

src/JSON.jl
```diff
@@ -68,116 +131,88 @@ function end_object(io::IO, state::State{NOINDENT}, is_dict::Bool)
...
-        for name in range[2:end]
-            Base.print(io, ",")
-            printsp(io, state)
-            Base.print(io, "\"", name, "\"", separator(state))
-            JSON._print(io, state, getfield(a, name))
-        end
+function _writejson(io::IO, state::State, a)
+    # FIXME: This fallback is harming performance substantially.
+    # Remove this fallback when _print removed.
+    if applicable(_print, io, state, a)
```
kmsquire: Since this call is costly, how about moving it to a global constant?

kmsquire:
Other than the one comment, I think this is good to merge.

README.md
```diff
@@ -67,3 +67,11 @@ provide a desired ordering.  For example, if you `import DataStructures`
...
 installed), you can pass `dicttype=DataStructures.OrderedDict` to
 maintain the insertion order of the items in the object.
+
+```julia
+JSON.lower(p::Point2D) = [x.x, x.y]
+```
+
+Define a custom serialization rule for a particular data type. Must return a
+value that can be directly serialized; see help for more details. Do **not**
+extend `JSON._print`; that is a clunky API and is now deprecated.
```
EricForgy: Maybe some nicer words could be chosen here?

> Please note that `JSON._print` is now deprecated in favor of `JSON.lower`.

TotalVerb: Fair enough. I do apologize for being so hostile to `JSON._print`; it was the only option, and this package's documentation is quite sparse, so from a consumer's viewpoint it's a very reasonable solution.

src/JSON.jl
TotalVerb: You mean to cache the results of the call? I'm a little worried about cache invalidation, but I suppose it's no worse than JuliaLang/julia#265 is.

TotalVerb: Caching the results in a `Dict` only saves a tiny portion of the runtime cost, while using more memory, adding compile-time overhead, and also making the entire thing not thread safe. I think we might have to live with the `applicable` check for now.

kmsquire: Ok, thanks for checking. One other idea to try would be to put a try-catch around a call to the _print function, temporarily replacing the io parameter with an IOBuffer to capture the output if the call succeeds. But I'm not expecting that to necessarily help.

TotalVerb: Using `try`-`catch` seems to make it even worse. I'll also try a `@noinline` function.

yuyichao: Just use `applicable` should be the best solution. It's doing a specialized dict lookup anyway. `try`-`catch` is also very expensive.

EricForgy:
Out of curiosity, why is deprecating a function that was never part of the API necessary? Especially if doing so has performance implications?

I overwrote `JSON._print` several times in several private repos, but I always understood it could break at some point because it wasn't part of the API. I suspect any other authors would have similar expectations and can adjust accordingly.

To be honest, looking at the code changes, I might be tempted to just find/replace `_print` -> `_writejson` :smile:

src/JSON.jl
TotalVerb: I meant a noinline function to remove the excess code generated by `depwarn`. I think that might help some, even if it doesn't remove the applicable cost.

TotalVerb:
Breaking four registered packages, one of which is `Plots`, is not an option.

Also, performance regressions will be fairly minor, and should disappear by next release.

src/JSON.jl
TotalVerb: Noinline function doesn't do much either. Anyway, the problem will be resolved when the deprecation warning is removed.

kmsquire: 👍 Should be good to merge.

src/JSON.jl
```diff
@@ -68,116 +131,88 @@ function end_object(io::IO, state::State{NOINDENT}, is_dict::Bool)
...
-function _print(io::IO, state::State, a::Associative)
+function _writejson(io::IO, state::State, a::Associative)
     if length(a) == 0
         Base.print(io, "{}")
         return
     end
     start_object(io, state, true)
     first = true
-    for (key, value) in a
+    for key in keys(a)
```
kmsquire: I just noticed this, and I'm curious about this change.  Is iterating with a tuple more expensive than computing a hash?

TotalVerb: Benchmarking shows that the tuple is cheaper (though it allocates three times the memory), but the cost of hashing is peanuts compared to the cost of IO in this loop, so the difference isn't detectable. I switched this because for `AssociativeWrapper`, the "hash" is just a fast `getfield` call; forming the tuple still has allocation issues. After #162, it will be possible to define separate serialization for `AssociativeWrapper` and `Dict`, so I switched back in that PR.

kmsquire: Okay, thanks for the clarification!  Not a big deal if it doesn't make much of a difference.  The larger memory allocation might be misleading, as it's likely that the allocated memory is being released/reused quickly.

PR #160, opened 2016-08-10
Add set of serialization microbenchmarks

TotalVerb:
These are the benchmarks that identified some major regressions in #151 (which have since been fixed). I'm pretty sure this file only works on v0.5, but I hope that's OK because as long as the package works on v0.3 and v0.4, I don't think it's critical to actively profile its performance.

### DIFF
--- a/bench/micro.jl
+++ b/bench/micro.jl
@@ -1,0 +1,54 @@
+# JSON Microbenchmarks
+
+using JSON
+using BenchmarkTools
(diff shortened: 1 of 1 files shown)


kmsquire:
👍

Should be merged at the same time as #151.

TotalVerb:
Going to merge this first so there's a easy way to benchmark the current master.

PR #174, opened 2016-10-02
Support for `Nullable`

### LINKED ISSUE #132: Write empty Nullable types as null
Apologies if I've misunderstood something, but it seems like Nullable values don't get written as JSON null if they are empty. Sorry for not setting up a pull request, but I think adding something like this should fix it:

    function _print{T}(io::IO, state::State, n::Nullable{T})
        if isnull(n)
            Base.print(io, "null")
        else
           _print(io, state, get(n))
        end
    end

joshbode:
Added `_writejson` method for `Nullable` values. Maps to JSON `null` if
value is null.

(edit by TotalVerb: fixes #132)

### DIFF
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -175,6 +175,10 @@ function _writejson(io::IO, state::State, a::Nullable)
     end
 end
 
+function _writejson(io::IO, state::State, c::Char)
+    _writejson(io, state, string(c))
+end
+
 function _writejson(io::IO, state::State, a::Associative)
     if length(a) == 0
         Base.print(io, "{}")
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -56,6 +56,7 @@ if VERSION < v"0.5.0-dev+2396"
     lower(f::Function) = "function at $(f.fptr)"
 end
 
+lower(c::Char) = string(c)
 lower(d::DataType) = string(d)
 lower(m::Module) = throw(ArgumentError("cannot serialize Module $m as JSON"))
 
@@ -173,10 +174,6 @@ function _writejson(io::IO, state::State, a::Nullable)
     else
         _writejson(io, state, get(a))
     end
-end
-
-function _writejson(io::IO, state::State, c::Char)
-    _writejson(io, state, string(c))
 end
 
 function _writejson(io::IO, state::State, a::Associative)
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -167,6 +167,14 @@ function _writejson(io::IO, state::State, n::Void)
     Base.print(io, "null")
 end
 
+function _writejson(io::IO, state::State, a::Nullable)
+    if isnull(a)
+        Base.print(io, "null")
+    else
+        _writejson(io, state, get(a))
+    end
+end
+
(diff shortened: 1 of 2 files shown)


joshbode:
Alternatively, `Char`s could be encoded as integers which has the nice property that they could be recovered directly via the constructor, e.g.

    Char(JSON.parse(JSON.json(x::Char)))

TotalVerb:
I think a string makes more sense for the default serialization, because most JSON is to communicate with other dynamic languages like JavaScript and Python, many of which treat their characters as string types instead of numeric ones. Packages will soon be able to define their own serializations anyway, which means we only need to provide a sensible default.

test/runtests.jl
```diff
@@ -251,6 +251,14 @@ end
 @test sprint(JSON.print, [NaN]) == "[null]"
 @test sprint(JSON.print, [Inf]) == "[null]"
 
+# check Nullables are printed correctly
+@test sprint(JSON.print, [Nullable{Int64}()]) == "[null]"
+@test sprint(JSON.print, [Nullable{Int64}(Int64(1))]) == "[1]"
+
+# check Chars
+@test json('a') == "\"a\""
+@test json('\\') == "\"\\\\\""
```
TotalVerb: Can we test the following too?

- a non-ASCII character
- a control character (like `\t`)

test/runtests.jl
```diff
@@ -251,6 +251,14 @@ end
 @test sprint(JSON.print, [NaN]) == "[null]"
 @test sprint(JSON.print, [Inf]) == "[null]"
 
+# check Nullables are printed correctly
+@test sprint(JSON.print, [Nullable{Int64}()]) == "[null]"
```
TotalVerb: Test `Nullable()` also

src/JSON.jl
```diff
@@ -167,6 +167,18 @@ function _writejson(io::IO, state::State, n::Void)
...
 
+function _writejson(io::IO, state::State, a::Nullable)
+    if isnull(a)
+        Base.print(io, "null")
+    else
+        _writejson(io, state, get(a))
+    end
+end
+
+function _writejson(io::IO, state::State, c::Char)
```
TotalVerb: This can probably be made a `lower` method with no performance penalty (since the string has to be allocated anyway). `JSON.lower(c::Char) = string(c)`. The `Nullable` one we can keep as-is, since `lower` would be type-unstable.

joshbode:
Thanks for the feedback :)

joshbode:
Is there anything else you need from me on this one?

TotalVerb:
@joshbode I was waiting on feedback from others, but I think this will be a pretty uncontroversial change. Will merge. Do you need it tagged?

joshbode:
Thanks for that! Yes please :)

TotalVerb:
Will do. Thank you for the contribution!

joshbode:
My pleasure - hoping to introduce JSON read/write for `DataFrame`s and it looks like they're migrating to `NullableArrays` (hence me wanting `Nullable` to be supported :)

TotalVerb:
https://github.com/JuliaLang/METADATA.jl/pull/6627

PR #175, opened 2016-10-05
adds lowering fallback for Real subtypes that converts to Float64

ssfrr:
My main use case for this is so that PlotlyJS.jl is able to plot
my vectors of fixed-point numbers, which come up a lot when loading
audio files.


stevengj:
LGTM.

ssfrr:
Thanks for the quick review and merge.

PR #178, opened 2016-12-02
Use Int instead of Int64

davidanthoff:
Otherwise any code that expects an ``Int`` on win32 will not work.

### DIFF
--- a/src/Parser.jl
+++ b/src/Parser.jl
@@ -322,9 +322,9 @@ the byte before `to`. Bytes enclosed should all be ASCII characters.
 """
 function int_from_bytes(bytes::Vector{UInt8}, from::Int, to::Int)
     @inbounds isnegative = bytes[from] == MINUS_SIGN ? (from += 1; true) : false
-    num = Int(0)
+    num = 0
     @inbounds for i in from:to
-        num = Int(10) * num + Int(bytes[i] - DIGIT_ZERO)
+        num = 10 * num + Int(bytes[i] - DIGIT_ZERO)
     end
     ifelse(isnegative, -num, num)
 end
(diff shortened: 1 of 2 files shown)


src/Parser.jl
```diff
@@ -322,9 +322,9 @@ the byte before `to`. Bytes enclosed should all be ASCII characters.
 """
 function int_from_bytes(bytes::Vector{UInt8}, from::Int, to::Int)
     @inbounds isnegative = bytes[from] == MINUS_SIGN ? (from += 1; true) : false
-    num = Int64(0)
+    num = Int(0)
```
stevengj: Why not just `num = 0`?  `Int` is the default type for Julia literals.

yuyichao:
Why would any code expect `Int32` if the code has always been like that?

src/Parser.jl
davidanthoff: Yes, you are right. Same is true two lines down, I guess, so I removed the ``Int`` there as well.

davidanthoff:
> Why would any code expect ``Int32`` if the code has always been like that?

My sense was always that one should use ``Int`` rather than ``Int64``. So that is what we did in our package, and then things break with the old version here. Could also well be that many folks never ran into this issue because few people actually run the 32 bit version of julia...

yuyichao:
Ref https://github.com/JuliaIO/JSON.jl/pull/140. Seems to be a intentional decision.

> My sense was always that one should use `Int` rather than `Int64`

This is definitely not true. For a machine independent storage format, it makes more sense if paring it on different machines returns the same result.

TotalVerb:
I doubt this will pass tests.

yuyichao:
The tests are changed so they can pass.

yuyichao:
See https://github.com/JuliaEditorSupport/LanguageServer.jl/blob/307771da964786796e841b424b3de91784d663d4/src/protocol.jl#L2, the constructor is over-tightening the input types.

davidanthoff:
@yuyichao Yes, that is the example. I guess we could just change the constructor to use ``Integer``... 

@TotalVerb That makes sense.

stevengj:
Presumably, the `Int64` was there for a reason.... if this is used for file lengths then it should be 64 bits even on 32-bit systems.

PR #180, opened 2017-01-04
compatibility for JuliaLang/julia#19449

stevengj:
In addition to removing some `string.data` constructs that will break after JuliaLang/julia#19449, I also removed a little pre-0.4 code (now obsolete since we REQUIRE 0.4).

### DIFF
--- a/src/specialized.jl
+++ b/src/specialized.jl
@@ -19,7 +19,7 @@ function parse_string(ps::MemoryParserState)
         parse_string(ps, b)
     end
 
-    _String(b)
+    Compat.UTF8String(b)
 end
 
 """
@@ -58,7 +58,7 @@ function predict_string(ps::MemoryParserState)
             if ps.utf8data[s] == LATIN_U  # Unicode escape
                 t = ps.s
                 ps.s = s + 1
-                len += length(string(read_unicode_escape!(ps)).data)
+                len += sizeof(string(read_unicode_escape!(ps)))
                 s = ps.s
                 ps.s = t
                 continue
@@ -97,7 +97,7 @@ function parse_string(ps::MemoryParserState, b)
             c = ps.utf8data[s]
             if c == LATIN_U  # Unicode escape
                 ps.s = s + 1
-                for c in string(read_unicode_escape!(ps)).data
+                for c in Vector{UInt8}(string(read_unicode_escape!(ps)))
                     i += 1
                     b[i] = c
                 end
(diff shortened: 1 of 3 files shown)


stevengj:
If you call the `codeunit` function, it should be as fast as accessing bytes, but less convenient.  Iterating over Unicode characters will always be slower that iterating over bytes, I expect.

stevengj:
(I don't have commit access, so someone else will have to click merge.)

src/specialized.jl
```diff
@@ -58,7 +58,7 @@ function predict_string(ps::MemoryParserState)
             if ps.utf8data[s] == LATIN_U  # Unicode escape
                 t = ps.s
                 ps.s = s + 1
-                len += length(string(read_unicode_escape!(ps)).data)
+                len += sizeof(string(read_unicode_escape!(ps)))
```
stevengj: This could certainly be made faster, since it is a bit silly to allocate a string just to compute its size.

```jl
utf8size(c::Char) = utf8size(UInt32(c))
utf8size(c::UInt32) = c < 0x80 ? 1 : c < 0x800 ? 2 : c < 0x10000 ? 3 : c < 0x110000 ? 4 : 3
```

src/specialized.jl
```diff
@@ -97,7 +97,7 @@ function parse_string(ps::MemoryParserState, b)
             c = ps.utf8data[s]
             if c == LATIN_U  # Unicode escape
                 ps.s = s + 1
-                for c in string(read_unicode_escape!(ps)).data
+                for c in Vector{UInt8}(string(read_unicode_escape!(ps)))
```
stevengj: could use the new `codeunit` function here to avoid allocating a `Vector{UInt8}` object.  Or we could re-implement [the Char-encoding logic](https://github.com/JuliaLang/julia/blob/887815c69d4511bb16f8b72a1a80bc57f97e044d/base/strings/string.jl#L186-L204) in order to avoid allocating a `String` at all.

src/specialized.jl
TotalVerb: The performance of Unicode escapes is a bit of a marginal use case, but you're right. It would be nice to have a package with functions like these for working with UTF8. (this also applies to the Char-encoding logic)

TotalVerb: I filed #181 to keep track of this issue.

src/specialized.jl
TotalVerb: We'll have to wait for `codeunit` to make its way into Compat first.

PR #182, opened 2017-01-04
Fix test deprecation warning on 0.6

TotalVerb:
Also bumps Compat dependency to <s>0.8.5</s> 0.9.5.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,2 @@
 julia 0.4
-Compat 0.8.4
+Compat 0.9.5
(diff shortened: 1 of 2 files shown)


REQUIRE
```diff
@@ -1,2 +1,2 @@
 julia 0.4
-Compat 0.8.4
+Compat 0.8.5
```
tkelman: take! was 0.9.5

TotalVerb: Thanks, fixed.

TotalVerb:
Restarting the failed travis job, just in case.

tkelman:
looked like a fluke download failure

TotalVerb:
Looks good now.

PR #184, opened 2017-01-18
support 0-dim arrays

alyst:
Currently, 0-dimensional arrays are not properly handled by the generic `_writejson()`.
The proposed specialized method just writes the first (and only) element.

### DIFF
--- a/src/Writer.jl
+++ b/src/Writer.jl
@@ -312,6 +312,9 @@ function show_json{T,n}(io::SC, s::CS, A::AbstractArray{T,n})
     end_array(io)
 end
 
+# special case for 0-dimensional arrays
+show_json{T}(io::SC, s::CS, A::AbstractArray{T,0}) = show_json(io, s, A[])
+
 show_json(io::SC, s::CS, a) = show_json(io, s, lower(a))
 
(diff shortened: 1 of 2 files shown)


TotalVerb:
Could you add a test?

TotalVerb:
It seems easier (and allocation-free) to write this as a `lower` method instead.

TotalVerb:
@alyst I see now why this might not work as a `lower` method, since `lower` wouldn't be called for any subtype of `AbstractArray`. Could you update this PR for the new `show_json` method?

TotalVerb:
In fact, the addition may be as simple as

```julia
function show_json{T,n}(io::SC, s::CS, A::AbstractArray{T,0})
    show_json(io, s, A[])
end
```

alyst:
@TotalVerb thanks for the suggestions. I've added the test and rebased the PR. I've also checked: if the `show_json(A::AbstractArray{T,0})` method is not defined, the new test is stuck at `newdims = ntuple(...)` of `show_json(A::AbstractArray{T,N})`. Would it make sense to add `@assert n > 0` there?

alyst:
travis failure seems unrelated

PR #185, opened 2017-01-19
Fix tests on 0.6

TotalVerb:
These two commits repair the tests on Julia v0.6.

### DIFF
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -17,7 +17,7 @@ import .Parser.parse
 "Internal JSON.jl implementation detail; do not depend on this type."
 immutable AssociativeWrapper{T} <: Associative{Symbol, Any}
     wrapped::T
-    fns::Array{Symbol, 1}
+    fns::Vector{Symbol}
 end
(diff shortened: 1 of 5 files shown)


test/REQUIRE
```diff
@@ -1,2 +1,2 @@
-DataStructures
+DataStructures 0.5.2
```
tkelman: are you using a feature or just relying on a bugfix? is the bugfix needed on all versions of julia that JSON supports, or just 0.6?

TotalVerb: The bugfix is needed on 0.6, but is not actually being used. I suppose that's not what these minimum version bounds are for. I'll remove this commit.

tkelman: if the package were 0.6-only, or there were a better way to say "apply this minimum package version for this version of julia" then it would make sense, but yeah for now maybe shouldn't restrict on 0.5 if not needed

TotalVerb:
CI passes. Good to go?

Issue #39, opened 2013-08-10
implement deserialize(T::Type, jsonstring::String)


stevengj:
Similar to Go's [Unmarshal](http://golang.org/pkg/encoding/json/#Unmarshal), as suggested in #38.

MrTomWhite:
Has there been any progress on this issue?

kmsquire:
I would guess that it has fallen off of people's radar.  Care to take a crack at it?

MrTomWhite:
I'll write up a pr soon. Thanks!

tlycken:
I have a local implementation of this. It's really simple:

```
using JSON

deserialize{T}(::Type{T}, json::AbstractString) = deserialize(T, JSON.parse(json))
deserialize{T}(::Type{T}, json::Dict) = T(flatten([deserialize(T, f, json) for f in fieldnames(T)])...)
deserialize{Tf}(::Type{Tf}, field::AbstractString, json::Dict) = deserialize(Tf, json[field])
deserialize{Tf<:Nullable}(::Type{Tf}, field::AbstractString, json::Dict) = haskey(json, field) ? deserialize(Tf, json[field]) : Tf()
deserialize{T}(::Type{T}, field::Symbol, json::Dict) = deserialize(fieldtype(T,field), string(field), json)
deserialize{T<:Integer}(::Type{T}, i::Integer) = T(i)
deserialize{T<:Integer}(::Type{T}, i::AbstractString) = parse(T, i)
deserialize{T<:AbstractFloat}(::Type{T}, i::AbstractFloat) = T(i)
deserialize{T<:AbstractFloat}(::Type{T}, i::AbstractString) = parse(T, i)
deserialize{T}(::Type{Nullable{T}}, x) = deserialize(T, x)

flatten{T<:Array}(A::Array{T,1}) = vcat(flatten(A)...)
flatten{T}(A::Array{T,1}) = vcat(A...)
flatten(A) = A
```

This works for a couple of cases that I could come up with, including making fields optional if their types are `Nullable`:

```
using Base.Test
input = "{ \"bar\": { \"baz\": 17 }, \"foo\": 3.14 }"

immutable Bar
    baz::Int
end

immutable Foo
    bar::Bar
end

immutable Baz
    foo::Nullable{Float64}
    bar::Bar
end

immutable Qux
    baz::Nullable{UTF8String}
    bar::Bar
end

@test deserialize(Foo, input) == Foo(Bar(17))
[...]

PR #186, opened 2017-01-20
Added attempt at unmarshal function

lwabeke:
First attempt at unmarshal function, along with a number of test cases
exercising the function with various combinations I could think of

### DIFF
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -271,4 +271,99 @@ function parsefile{T<:Associative}(filename::AbstractString; dicttype::Type{T}=D
     end
 end
 
+export unmarshal # returns a reconstrcuted variable from a JSON parsed string
+"""
+Called with a given DataType and using the dictionary output of a JSON.parse , it will try to reconstruct the DataType from the JSON dictionary.
+
+    unmarshal(typeof(var), JSON.parse(JSON.json(var)) == var
+"""
+function unmarshal{T}(DT :: Type{Array{T}}, parsedJson)
+    #    @show "unmarshalArrayComposite"
+
+    tmp = DT(length(parsedJson))
+    for cnt=1:length(parsedJson)
+        tmp[cnt] = unmarshal(T, parsedJson[cnt])
+    end
+    tmp
+end
+
+
+function unmarshal{T <: Number}(DT :: Type{Array{T}}, parsedJson)
+#    @show "unmarshalArrayNumber"
+    map(T, parsedJson)
+end
+
+function unmarshal{T <: Number}(DT :: Type{Array{Complex{T}}}, parsedJson)
+#        @show "unmarshalArrayComplex"
+    tmp = DT(length(parsedJson))
+    for cnt=1:length(parsedJson)
+        tmp[cnt] = unmarshal(Complex{T}, parsedJson[cnt])
+    end
+    tmp    
+end
+
+
+function unmarshal{T}(DT :: Type{Array{T, 2}}, parsedJson)
+#    @show "unmarshalArray2"
+
+    sizes = (length(parsedJson[1]), length(parsedJson) )
+    tmp = DT(sizes)
+    for cnt=1:length(parsedJson)
+        tmp[:, cnt] = unmarshal(Array{T}, parsedJson[cnt])
+    end
+        
+    tmp
+end
+
+
+function unmarshal{T, N}(DT :: Type{Array{T, N}}, parsedJson)
+#    @show "unmarshalArrayN"
+    if N == 1
+        # This gets called if type was defined as Array{T,1}, instead of Array{T}
+        return unmarshal(Array{T}, parsedJson)
+    end
+
+    sizes = (length(parsedJson) )
+    p = parsedJson[1]
+    for cnt=2:N
+        sizes = (length(p), sizes... )
+        p = p[1]
+        end
+#    @show sizes
+    tmp = DT(sizes...)
+    subblocksize = prod(sizes[1:N-1])
+    for cnt=1:length(parsedJson)
+        tmp[(1+(cnt-1)*subblocksize):(cnt*subblocksize)] = unmarshal(Array{T, N-1}, parsedJson[cnt])
+    end
+        
+    tmp
+end
+
+
+function unmarshal(DT, parsedJson)
+#    @show "unmarshalStruct"
+
+    tup = ()
+    for iter in fieldnames(DT)
+        DTNext = fieldtype(DT,iter)
+#        @show iter, DTNext, !haskey(parsedJson, string(iter)) 
+
+        if !haskey(parsedJson, string(iter)) 
+            val = DTNext()
+        else
+            if (DTNext <: Array) || (typeof(parsedJson[string(iter)]) <: Dict{String,Any})
+                val = unmarshal( DTNext, parsedJson[string(iter)])
+            else
+                val = DTNext(parsedJson[string(iter)])
+            end
+        end
+            
+        tup = (tup..., val)
+    end
+    
+    DT(tup...)
+end
+
+
+
 end # module
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -349,12 +349,27 @@ function unmarshal(DT, parsedJson)
 #        @show iter, DTNext, !haskey(parsedJson, string(iter)) 
 
         if !haskey(parsedJson, string(iter)) 
-            val = DTNext()
+            try
+            	val = DTNext()
+            catch ex
+            	if isa(ex, MethodError)
+			error("Key $(string(iter)) is missing from the structure $(DT) and field is not Nullable")
+                end
+                rethrow(ex)
+            end # try-cath
         else
             if (DTNext <: Array) || (typeof(parsedJson[string(iter)]) <: Dict{String,Any})
                 val = unmarshal( DTNext, parsedJson[string(iter)])
             else
-                val = DTNext(parsedJson[string(iter)])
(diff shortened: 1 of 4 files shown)


lwabeke:
Attempts to address: 
https://github.com/JuliaIO/JSON.jl/issues/39

TotalVerb:
Is this JSON-specific? It seems it could be part of a separate package, as it could be useful for similar formats (e.g. CSV, CSON, TOML, etc.) in greater generality.

lwabeke:
I guess it could be separate, but makes the assumption that the format has already been parsed and converted into a hierarchical structure consisting of dictionaries and arrays. A 1 minute google on CSON and TOML looks like they might fit the concept, if there is already a parser that meets that requirement (other than the 1 minute googling I have no experience of these and don't know finer implications).
On CSV I'm not sure: Typically it is not hierarchical, rather thus a table format. While I guess it could work in principle, there are probably ways to get it a much more efficient implementation, by taking that fact into consideration.

TotalVerb:
There is [TOML.jl](https://github.com/pygy/TOML.jl) which parses into a similar format as JSON.jl. I don't know if there is a CSON parser yet.

You're right about CSV; this would probably only be useful for JSON-like file formats. But there seem to be plenty of those to go around.

lwabeke:
After submitting, I encountered a fundamental problem, but I suspect it is due to Julia Language itself, and possibly unsolvable:

```
using JSON

type Reconfigurable{T}
    x :: T
    y :: T
    z :: Int
end


val = Reconfigurable(1.0, 2.0, 3)
jstring = JSON.json(val)
JSON.unmarshal(Reconfigurable{Float64}, JSON.parse())
```
This work fine if we manually fully specify the parameteric type.

However fails if we call it with the specification:

```
JSON.unmarshal(reconfigurable, JSON.parse())
```
My calls to fieldtype on reconfigurable's fields just returns T, which is understandable I guess.

The problem is if it is not your own type definition, you as user might not be aware of the need to specify the fieldtypes to make it concrete. And/or the sequence to use if multiple fields are left parametric. Even a nice warning/error return is difficult, since it is not easy to detect it. Code can check for T and hope that is frequently used as a placeholder, but package developers could end up using other placeholders...

While the user can still work around the issue here, the problem becomes worse when reusing packages that were developed to be generic and which does the creation of these structures internally in a hierarchical way:


```
type Higherlayer
    val :: Reconfigurable
end

higher = Higherlayer(val)
jstring = JSON.json(higher)
JSON.unmarshal(Higherlayer, JSON.parse(jstring))
```

As far as I can see, there is no way to fully define the type of Higherlayer as 
Higherlayer{Reconfigurable{Float64}}, unless we go and modify its definition (which if inside a package we are using, can become unmanagable). And we also don't want to break this feature of generic code by forcing all types to be fully defined. 

While I don't think these limitations makes the inclusion of this feature unusable, it does place restrictions on its use, which could be difficult to fully explain to the average user.

Any thoughts?

TotalVerb:
I have thought about this problem in the past. It is late here; tomorrow I will take some time to explain my thoughts.

TotalVerb:
@lwabeke My idea is, simply, to design `unmarshal` as a function that has a sensible default for fully-specified concrete types, but to note define it at all for abstract types. Instead, the user is encouraged to overload `unmarshal` for non-concrete types. The contract is that, `unmarshal(T, x)` must return a value of type `T`.

For instance, if `unmarshal(Reconfigurable, x)` is called, the user will implement this function and determine an appropriate type parameter `T` to return a value of `Reconfigurable{T}`. Thus the construction of `Higherlayer` objects can still use the default behaviour, which automatically calls the user-configured behaviour.

lwabeke:
I figured out how to try catch some of the errors related to these abstract types and updated the code with easier to understand error messages, compared to the defaults that gets invoked in this case.

src/JSON.jl
```diff
@@ -271,4 +271,114 @@ function parsefile{T<:Associative}(filename::AbstractString; dicttype::Type{T}=D
...
+    subblocksize = prod(sizes[1:N-1])
+    for cnt=1:length(parsedJson)
+        tmp[(1+(cnt-1)*subblocksize):(cnt*subblocksize)] = unmarshal(Array{T, N-1}, parsedJson[cnt])
+    end
+        
+    tmp
+end
+
+
+function unmarshal(DT, parsedJson)
```
TotalVerb: I would write `unmarshal(T::DataType, parsedJson)`, because we need `fieldnames`, which is not available for other types (e.g. `Union{Int,String}`)

src/JSON.jl
```diff
@@ -271,4 +271,114 @@ function parsefile{T<:Associative}(filename::AbstractString; dicttype::Type{T}=D
...
+#    @show "unmarshalStruct"
+
```
TotalVerb: Should check `isleaftype(T)`; otherwise throw ArgumentError; can't default-unmarshal an abstract type.

src/JSON.jl
```diff
@@ -271,4 +271,114 @@ function parsefile{T<:Associative}(filename::AbstractString; dicttype::Type{T}=D
...
+    tup = ()
+    for iter in fieldnames(DT)
+        DTNext = fieldtype(DT,iter)
+#        @show iter, DTNext, !haskey(parsedJson, string(iter)) 
+
+        if !haskey(parsedJson, string(iter)) 
+            try
+            	val = DTNext()
```
TotalVerb: Is there a reason why the existence of a zero-argument constructor is tied to whether a field can be omitted? I think it would be more sensible to define a `default(...)` function, and define it only for things like `Nullable`.

src/JSON.jl
```diff
@@ -271,4 +271,114 @@ function parsefile{T<:Associative}(filename::AbstractString; dicttype::Type{T}=D
...
+            catch ex
+            	if isa(ex, MethodError)
+			error("Key $(string(iter)) is missing from the structure $(DT) and field is not Nullable")
```
TotalVerb: watch the whitespace

src/JSON.jl
```diff
@@ -271,4 +271,114 @@ function parsefile{T<:Associative}(filename::AbstractString; dicttype::Type{T}=D
...
+                end
+                rethrow(ex)
+            end # try-cath
+        else
+            if (DTNext <: Array) || (typeof(parsedJson[string(iter)]) <: Dict{String,Any})
```
TotalVerb: I think the only sensible way is to recursively call unmarshal always, otherwise this will not work for nested JSON objects. That is, delete the else branch here.

src/JSON.jl
```diff
@@ -271,4 +271,114 @@ function parsefile{T<:Associative}(filename::AbstractString; dicttype::Type{T}=D
...
+                val = unmarshal( DTNext, parsedJson[string(iter)])
+            else
+        #      @show parsedJson[string(iter)] 
+                try
+                	val = DTNext(parsedJson[string(iter)])
+                catch ex
+                	if isa(ex, UndefVarError) || isa(ex, MethodError)
```
TotalVerb: how can `UndefVarError` be thrown here?

src/JSON.jl
```diff
@@ -271,4 +271,114 @@ function parsefile{T<:Associative}(filename::AbstractString; dicttype::Type{T}=D
...
+            end
+        end
+            
+        tup = (tup..., val)
+    end
+    
+    DT(tup...)
+end
+
+
```
TotalVerb: finally, I'd suggest a series of `unmarshal` overloads like

```julia
unmarshal{T<:Number}(::Type{T}, x::Number) = T(x)
```

to deal with the "base cases" of unmarshaling.

TotalVerb:
Thanks for this contribution; I've reviewed parts of it. I still think that it would make sense as a separate package.

src/JSON.jl
lwabeke: My reasoning was that Nullables and any type that can construct from a zero-argument constructor could be used to complete the creation. I guess we can constrict it to only construct Nullables if missing and assume of all other types it needs to be explicitly included in the JSON string.
I would be hesitant to on my own introduce a `default(...)` function and then expect all the types across the whole Julia ecosystem to consider having to add that function.

src/JSON.jl
lwabeke: Not sure I understand the working of isleaftype(T) properly, but if I understand correct, I believe that is what I am trying to do with the later catch containing:
                               ` error("Cannot construct $(DTNext)($(parsedJson[string(iter)])). Type $(DT) might be not be concrete. Consider implementing overload function unmarshal(T ::$(DT), dict) to return a value of type $(DT)")`

Should I rather use `isleaftype() `instead of the exception handling? But if I do, I would also need to check if the user has supplied the necessary method. How would I go about doing that?
`methods()` ? Looks like it might be able to work, but not sure how to parse its output.

lwabeke:
I have thought hard about your suggestion about a separate package and yes I think I agree.
If it can be panelbeated to be able to support the other formats as you suggested, then having it external would be the best, instead of creating similar functions in each of the packages. 
A first aim would then just be to get it setup and running for JSON, but with the intention to try supporting some of the others as a next iteration (maybe I'm lucky and it works out of the box, but I can think there might be some tweaks required on some formats).

I have started reading up on package creation and will try to move this across. I'm think of an unmarshal package.

lwabeke:
Package Unmarshal created. Awaiting merge into METADATA list:

https://github.com/lwabeke/METADATA.jl/compare/pull-request/c487d969


Pkg.clone(https://github.com/lwabeke/Unmarshal.jl.git)

PR #200, opened 2017-04-24
Drop v0.4 support

TotalVerb:
This PR drops 0.4 support. I don't know if it's the best time to do so; feedback is welcome.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ os:
     - osx
     - linux
 julia:
-    - 0.4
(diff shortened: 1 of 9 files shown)


tkelman:
If we do https://github.com/JuliaLang/METADATA.jl/pull/9108 then this could even be point-releaseable

TotalVerb:
No point waiting on this now with the METADATA PR done. But there's also no rush to release, since there should be no change at all from the last one.

PR #202, opened 2017-04-27
Support Enums

### LINKED ISSUE #201: Support enums
I noticed that enums are not supported currently, but only show up like
```julia
julia> JSON.json(r)
"{\"exprType\":\"DATE_GRAIN_COLUMN\",\"dateTimeElement\":{},\"interval\":1}"
```
where the Julia object is
```julia
julia> r = ans
InsightsApp.DQL.DateGrainColumn("DATE_GRAIN_COLUMN", DAY::InsightsApp.DQL.DateTimeElement = 3, 1)
```

TotalVerb:
Fixes #201. I made some unrelated changes also:

- The `JSONPrimitive` constant is not used anywhere and so I've deleted it.
- To avoid excess allocations in lowering `Char` and `Type`, they're now lowered to themselves and printed directly by the string-like `show_json` (previously, we were doing that only for symbols and strings). The same treatment applies to `Enum`; users can still define `lower` to customize the printing if necessary (see test/lowering.jl). We can do this with dates too, but that's easier to do once 0.4 support is dropped.

### DIFF
--- a/src/Writer.jl
+++ b/src/Writer.jl
@@ -17,18 +17,15 @@ immutable CompositeTypeWrapper{T}
 end
 CompositeTypeWrapper(x) = CompositeTypeWrapper(x, fieldnames(x))
 
-const JSONPrimitive = Union{
-        Associative, Tuple, AbstractArray, AbstractString, Integer,
-        AbstractFloat, Void, CompositeTypeWrapper}
-
 """
     lower(x)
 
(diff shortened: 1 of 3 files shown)


TotalVerb:
@quinnj, could you have a look at this?

quinnj:
Looks good to me!

PR #204, opened 2017-05-08
Modernize test suite for JSON

TotalVerb:
This reshuffles the JSON test code to use `@testset` and to be more organized in terms of files. The only test that was actually changed was the dicttype test, which is written in a more robust way now (the previous test wasn't testing much).


TotalVerb:
If there are no objections, comments, or requests for time, I will merge this on Wednesday.

Issue #203, opened 2017-05-07
Cannot serialize type DataArrays.NAtype


disleyland:
JSON.json is failing when there are NAs in a string data array

TotalVerb:
What is the expected behavior? `null`?

quinnj:
That's what I would expect.

TotalVerb:
This can't be fixed in JSON.jl because we can't define anything on `DataArrays.NAtype` without requiring it, and JSON is the more fundamental package here. Please raise an issue on `DataArrays`.

nalimilan:
What kind of methods should be defined in DataArrays to fix this? Cannot JSON.jl simply call `string`, which would return `"NA"`?

TotalVerb:
To be consistent with `Nullable`, we shouldn't use `"NA"` but rather `null` as the JSON equivalent. JSON does not try to guess at which objects want to be serialized as strings. We could serialize all singleton objects as strings, but that behavior seems unintuitive to me. Alternatively, we could serialize all singleton objects as `null`, but that also sounds suspicious. The best solution (as is being done right now) is probably to simply fail and leave it up to the package author.

The method that would need to be defined is

```
JSON.lower(::NAtype) = nothing
```

TotalVerb:
I'm not opposed to a blanket rule saying "serialize all singleton objects as _______" either: but we would need to bikeshed whether this should be `string`, `null`, or something else.

nalimilan:
I really don't know, and the goal is to move to a common representation for missingness quite soon anyway. But clearly we don't want DataArrays to depend on JSON, and for now we don't have optional modules.

TotalVerb:
I wonder if it would make sense to move JSON.lower to a minimal `IOBase` package that just serves as a namespace to contain `lower` and related functions. Then we can avoid a JSON dependency; an IOBase one probably isn't so bad?

disleyland:
speaking from the perspective that when I use JSON.jl, the file is to be processed by scripts in a browser, the old julia position was odd because it treated missingness as {} in the output string, which then got interpreted as an blank object in a browser based javascripting environment.  in the browser, to then check if this object was truly empty or hiding some other useful information, we'd need to have further checks which is a coding annoyance when trying to build re-usable code for lots of JSON outputs from Julia.  Seems very sensible for JSON.jl to treat NA as nothing (is that different then null?) as the resulting string is unambiguous when processed in a browser

nalimilan:
[...]

PR #205, opened 2017-05-10
RFC (don't merge) Send every singleton to nothing

TotalVerb:
This is a possible solution to #203. I don't particularly like this solution; I think it should not be JSON's responsibility to guess that all singletons should turn into `null`, but I can't think of a better one for #203.

### DIFF
--- a/src/Writer.jl
+++ b/src/Writer.jl
@@ -38,9 +38,8 @@ primitives.
 function lower(a)
     if nfields(typeof(a)) > 0
         CompositeTypeWrapper(a)
-    else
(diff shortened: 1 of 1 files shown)


TotalVerb:
cc @disleyland

TotalVerb:
My personal preference is still to move `JSON.lower` to a more fundamental package, say `IOBase`, and have `DataArrays` depend on and extend that.

TotalVerb:
It honestly does not seem to me like this is a good idea, and there has not been so much demand so far. Closing.

PR #206, opened 2017-05-13
Lower `AbstractSet` through `collect`

### LINKED ISSUE #196: Serialization of Set should (probably) not dump the current dictionary representation
This may have to be changed in the Set implementation instead, current behaviour : 
`julia> VERSION`
`v"0.5.1"`
`julia> json( Set([1,2]) )`
`"{\"dict\":{\"2\":null,\"1\":null}}"`

probably desired behaviour:
`julia> json( collect(Set([1,2])) )`
` "[2,1]" `
or define:
`JSON.lower(s::Set) = collect(s)`
since Set is a basic type, this might break some code (I doubt anyone counts on this behaviour)

TotalVerb:
Fix #196.

This has been requested for a while and makes sense to do. The first commit is unrelated but just a simplification to the code with no functional change.

@oyd11, please review and confirm that this addresses your use case.

### DIFF
--- a/src/Writer.jl
+++ b/src/Writer.jl
@@ -43,14 +43,13 @@ function lower(a)
     end
 end
 
-lower(s::Base.Dates.TimeType) = string(s)
-
 # To avoid allocating an intermediate string, we directly define `show_json`
 # for this type instead of lowering it to a string first (which would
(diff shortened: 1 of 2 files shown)


TotalVerb:
Unless there are objections, comments, or requests for time, I'll merge and tag this on Monday.

PR #208, opened 2017-05-18
Fix the deprecation warnings

TotalVerb:
Since benchmarks are not a part of JSON itself, I didn't bother trying to support both 0.5 and 0.6 on them; this PR makes 0.6 is required to run the benchmarks.

This is a really ugly PR but luckily only a temporary solution until we can drop support for 0.5.

### DIFF
--- a/bench/micro.jl
+++ b/bench/micro.jl
@@ -1,4 +1,5 @@
 # JSON Microbenchmarks
+# 0.6 required for running benchmarks
 
 using JSON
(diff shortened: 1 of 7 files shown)


TotalVerb:
Modulo comments, objections, or requests for time, I will merge this on Wednesday.

tkelman:
I'd have either made master 0.6-only for this or not worried about it for now, since this might be fragile too and is pretty ugly.

PR #209, opened 2017-05-19
Support Pair, serializing as an object

### LINKED ISSUE #106: JSON.json(::Pair)
Could we have the following?

``` julia
t = Pair("hey",1)
JSON.json(t)
```

results in

``` json
{\"hey\": 1}
```

or would this go against a fundamental tenet?

TotalVerb:
@quinnj, @kmsquire please review if you have the time. I implemented this as `show_json` instead of `lower` because allocating `Dict`s is expensive...

Close #106.

### DIFF
--- a/src/Writer.jl
+++ b/src/Writer.jl
@@ -282,6 +282,12 @@ function show_json(io::SC, s::CS, a::Associative)
     end_object(io)
 end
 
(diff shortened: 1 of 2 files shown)


quinnj:
Just a quick question: will this always set the key as a JSON string?

TotalVerb:
Yes.

PR #215, opened 2017-09-12
Fix loading depwarn on 0.7

yuyichao:


yuyichao:
Added another fix for loop variable scope

PR #217, opened 2017-09-23
Drop v0.5 support, Compat dependency

TotalVerb:
This drops the 0.5 support for JSON, and along with it several hacks and Compats we've been using. Compat is still needed for tests. Also has the effect of fixing all the deprecations on master.

Will merge and tag in a few days if there are no objections.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - osx
     - linux
 julia:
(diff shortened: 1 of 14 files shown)


tkelman:
You can't remove Compat, you still use `read(io, String)` in src

PR #219, opened 2017-09-25
Revert Compat dependency removal

TotalVerb:
cc @tkelman 

Sorry for the oversight. This partially reverts the removal of the Compat dependency in 0.15.0. Should be tagged as 0.15.1 when merged.

(as an aside: for these reasons, I suppose it's best for tests to avoid importing Compat when the main library doesn't, to avoid these situations being missed.)

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.6
+Compat 0.30.0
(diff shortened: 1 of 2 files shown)


TotalVerb:
The deletion of test/REQUIRE was accidental, sorry. Fixed in the latest force commit.

TotalVerb:
Going to merge and tag this hotfix.

PR #224, opened 2017-11-17
Allow user to specify integer parsing type (fixes #223)

### LINKED ISSUE #223: Large unsigned integer returns nonsensical result
Is it possible to define the type of the dict in the parsing command.
My json looks like 
`{"201736327611975630":18005722827070440994}` which seems to be handled as String and then kind of integer but it's actually unsigned. Therefore after parsing I get results ending in something like `-441021246639110622`

kmsquire:
* Added a ParserContext object, containing the `DictType` and `IntType` to use for parsing JSON

This is a minimally invasive way of giving the user more control over parsing.  

A possible extension would be to allow specifying the float format to parse (e.g., `DecFP`, `Decimals`, `FixedPointNumbers`, etc.), although this would probably require a bit more work.

**Edit:** Addresses part of #223, although it does not implement overflow checking.

### DIFF
--- a/src/Parser.jl
+++ b/src/Parser.jl
@@ -30,6 +30,8 @@ mutable struct StreamingParserState{T <: IO} <: ParserState
 end
 StreamingParserState(io::IO) = StreamingParserState(io, 0x00, true)
 
+struct ParserContext{DictType, IntType} end
+
 """
 Return the byte at the current position of the `ParserState`. If there is no
 byte (that is, the `ParserState` is done), then an error is thrown that the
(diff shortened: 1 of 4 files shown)


src/Parser.jl
```diff
@@ -197,9 +199,9 @@ function parse_array(ps::ParserState, dictT::Type)
 end
 
 
-function parse_object(ps::ParserState, dictT::Type)
-    obj = dictT()
-    keyT = dictT.parameters[1]
+function parse_object(ps::ParserState, pc::ParserContext{DictType, <:Real}) where DictType
+    obj = DictType()
+    keyT = DictType.parameters[1]
```
TotalVerb: perhaps `keytype(...)`, although there's probably a reason we aren't already using it

test/parser/inttype.jl
```diff
@@ -0,0 +1,16 @@
+@testset for T in [Int32, Int64, Int128]
+  val = JSON.parse("{\"x\": 3}", inttype=T)
```
TotalVerb: we usually use 4 space indents

TotalVerb:
:+1:, I think this is a good idea.

test/parser/inttype.jl
kmsquire: Right, fixed in next push

src/Parser.jl
kmsquire: Yes, `keytype(...)` would be better.  I suspect the code was originally written before that function existed.

kmsquire:
Thanks, @TotalVerb, I made the suggested changes, updated the README, and made some other cosmetic changes.  Because of the additional, let me know if you think anything else needs to change.  Otherwise, this can be squashed and merged.

TotalVerb:
Looks great, thanks!

Issue #225, opened 2017-11-22
Movement of Dates on Nightly


Nosferican:
Since Julia [0c0d6cc](https://github.com/JuliaLang/julia/commit/0c0d6cca5d3793eaf6321ef0577afd6ce594a9b5) JSON fails to compile due to `Writer.jl`
```
const IsPrintedAsString = Union{
    Dates.TimeType, Char, Type, AbstractString, Enum, Symbol}
```
which gives
```
ERROR: LoadError: LoadError: type Void has no field TimeType
```
Fix is to add the line to `Writer.jl`
```
import Dates
```

KristofferC:
```
if VERSION >= v"0.7.0-DEV.2575"
    import Dates
end
```

Nosferican:
Thanks!

TotalVerb:
Closed by #226.

PR #226, opened 2017-11-22
Address breaking change on nightly

Nosferican:
Fix to [225](https://github.com/JuliaIO/JSON.jl/issues/225)

### DIFF
--- a/src/Writer.jl
+++ b/src/Writer.jl
@@ -1,5 +1,6 @@
 module Writer
 
+import Dates
 using ..Common
 using ..Serializations: Serialization, StandardSerialization,
                         CommonSerialization
--- a/src/Writer.jl
+++ b/src/Writer.jl
@@ -1,8 +1,6 @@
 module Writer
 
-if VERSION >= v"0.7.0-DEV.2575"
-    import Dates
-end
+using Compat
 using ..Common
 using ..Serializations: Serialization, StandardSerialization,
                         CommonSerialization
--- a/src/Writer.jl
+++ b/src/Writer.jl
@@ -1,6 +1,8 @@
 module Writer
 
-import Dates
+if VERSION >= v"0.7.0-DEV.2575"
+    import Dates
+end
 using ..Common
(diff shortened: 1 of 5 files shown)


TotalVerb:
Let's wait for Compat to support this so this can be done as `using Compat.Dates`.

TotalVerb:
ref https://github.com/JuliaLang/Compat.jl/pull/413

andreasnoack:
The new version of Compat is now in METADATA so this can be updated

src/Writer.jl
```diff
@@ -1,5 +1,6 @@
 module Writer
 
+using Compat
```
andreasnoack: I think you also need a `using Compat.Dates`

Nosferican: @nalimilan is the `using Compat.Dates` required as well?

nalimilan: Yes, you need to replace `Base.Dates` with `Compat.Dates` everywhere it appeared. So I now realize https://github.com/JuliaData/Missings.jl/pull/56 didn't change anything to the problem.

nalimilan: Regarding the case at hand, you just need to replace `Dates` with `Compat.Dates` on this line: https://github.com/JuliaIO/JSON.jl/blob/b4a4e338029bb999589dfabc6d90f5ae2fdd08d8/src/Writer.jl#L52

Nosferican: Let us check change it and check Travis. After the manual fix both packages compiled so the CI should be cleared.

stevengj:
There's no performance difference between `using Dates` and `import Dates` (followed by qualified names).  As long as there are no namespace collisions, I think `using Dates` is fine.

TotalVerb:
Hi @Nosferican, thank you for the contribution! I went ahead and removed the `if isdefined(...)` line as mentioned in a review above. I agree that importing/using Dates is not a big deal, so this LGTM once CI passes.

PR #228, opened 2017-12-04
Make it possible to specify float type as well

Keno:
I have an application dealing with decimal money amounts sent via JSON,
I want to parse these as a FixedPointDecimal (
https://github.com/JuliaMath/FixedPointDecimals.jl) to make sure floating
point doesn't mess up my amounts. This code is sufficient to convience
JSON.jl to parse numbers as FixedPointDecimals.

### DIFF
--- a/src/Parser.jl
+++ b/src/Parser.jl
@@ -31,7 +31,7 @@ mutable struct StreamingParserState{T <: IO} <: ParserState
 end
 StreamingParserState(io::IO) = StreamingParserState(io, 0x00, true)
 
-struct ParserContext{DictType, IntType} end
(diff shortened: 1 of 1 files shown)


kmsquire:
Looks good!  Can you add some tests?

ScottPJones:
It would be good to have this tested with https://github.com/stevengj/DecFP.jl as well.

kmsquire:
@Keno are you still interested in this pull request?

Keno:
abstractly yes, but I'm unlikely to have the time to get back to it in the near future, so feel free to close.

PR #229, opened 2017-12-13
Fix for Base Unicode functions moving to stdlib

ararslan:
Ref https://github.com/JuliaLang/julia/pull/25021. Needed for Nanosoldier.


ararslan:
Needs https://github.com/JuliaLang/METADATA.jl/pull/12476

ararslan:
Oh, yeah I guess so.

PR #230, opened 2017-12-14
Associative -> AbstractDict

ararslan:


TotalVerb:
This change should go in Compat first.

Issue #232, opened 2018-01-19
JSON.parse fails for cweb/url-testing/master/urls.json


samoconnor:
Input: https://raw.githubusercontent.com/cweb/url-testing/master/urls.json

```julia
julia> r = HTTP.get("https://raw.githubusercontent.com/cweb/url-testing/master/urls.json")
julia> JSON.parse(HTTP.payload(r, String))
ERROR: Expected 'u' here
Line: 36
Around: ...\\\ud800\\\u597D",          "e...
                    ^

Stacktrace:
 [1] _error(::String, ::JSON.Parser.MemoryParserState) at /Users/sam/.julia/v0.6/JSON/src/Parser.jl:135
 [2] skip! at /Users/sam/.julia/v0.6/JSON/src/Parser.jl:76 [inlined]
 [3] read_unicode_escape!(::JSON.Parser.MemoryParserState) at /Users/sam/.julia/v0.6/JSON/src/Parser.jl:257
 [4] predict_string(::JSON.Parser.MemoryParserState) at /Users/sam/.julia/v0.6/JSON/src/specialized.jl:60
 [5] parse_string(::JSON.Parser.MemoryParserState) at /Users/sam/.julia/v0.6/JSON/src/specialized.jl:5
 [6] parse_value(::JSON.Parser.ParserContext{Dict{String,Any},Int64}, ::JSON.Parser.MemoryParserState) at /Users/sam/.julia/v0.6/JSON/src/Parser.jl:157
 [7] parse_object(::JSON.Parser.ParserContext{Dict{String,Any},Int64}, ::JSON.Parser.MemoryParserState) at /Users/sam/.julia/v0.6/JSON/src/Parser.jl:218
 [8] parse_value(::JSON.Parser.ParserContext{Dict{String,Any},Int64}, ::JSON.Parser.MemoryParserState) at /Users/sam/.julia/v0.6/JSON/src/Parser.jl:161
 [9] parse_array(::JSON.Parser.ParserContext{Dict{String,Any},Int64}, ::JSON.Parser.MemoryParserState) at /Users/sam/.julia/v0.6/JSON/src/Parser.jl:191
 [10] parse_value(::JSON.Parser.ParserContext{Dict{String,Any},Int64}, ::JSON.Parser.MemoryParserState) at /Users/sam/.julia/v0.6/JSON/src/Parser.jl:163
[...]

PR #234, opened 2018-01-24
Fix deprecations with copy! and Vector{UInt8}(str)

ScottPJones:


src/Parser.jl
```diff
@@ -272,7 +273,7 @@ function parse_string(ps::ParserState)
         if c == BACKSLASH
             c = advance!(ps)
             if c == LATIN_U  # Unicode escape
-                append!(b, Vector{UInt8}(string(read_unicode_escape!(ps))))
+                append!(b, get_bytes(string(read_unicode_escape!(ps))))
```
stevengj: `append!(b, codeunits(s))` would work here … no need to force a copy to be made as in `get_bytes`

src/Parser.jl
```diff
@@ -381,9 +382,11 @@ function unparameterize_type(T::Type)
     candidate <: Union{} ? T : candidate
 end
 
-function parse(str::AbstractString; dicttype::Type{<:AbstractDict}=Dict{String,Any}, inttype::Type{<:Real}=Int64)
+function parse(str::AbstractString;
+               dicttype::Type{<:AbstractDict}=Dict{String,Any},
+               inttype::Type{<:Real}=Int64)
     pc = ParserContext{unparameterize_type(dicttype), inttype}()
-    ps = MemoryParserState(Vector{UInt8}(String(str)), 1)
+    ps = MemoryParserState(get_bytes(str), 1)
```
stevengj: It would be better not to make a copy here.

Since `MemoryParserState` is only ever created from a `String`, it would be better to just (a) change the `utf8data` field to a `String` field and (b) call `codeunit(s, i)` to extract a byte from the string (e.g. in the `byteat` function).  Note that `codeunit` already works in Julia 0.6.

src/Parser.jl
```diff
@@ -262,6 +262,7 @@ function read_unicode_escape!(ps)
     end
 end
 
+get_bytes(str) = Vector{UInt8}(@static isdefined(Base, :codeunits) ? codeunits(str) : str)
```
stevengj: This function is not necessary given the changes I suggest below.

However, it would be good to have a `Compat.codeunits(s)` function that just returns `Vector{UInt8}(s)` in Julia 0.6.

stevengj: Just pushed a PR for Compat to add `codeunits` and `ncodeunits`

src/Parser.jl
ScottPJones: The problem was that `codeunits` wasn't defined in `Compat.jl` to make it work on v0.6.

src/Parser.jl
ScottPJones: Yes, but I'm trying to make the minimal changes here (as I was told many times before to do 🙂 ) to make it work on both v0.6 and later.

src/Parser.jl
ScottPJones: If those are merged soon, then yes, I'll change this to use codeunits.

src/Parser.jl
stevengj: My suggestions require hardly any more code, and don't make more copies.

ScottPJones:
I tried to put this bump message on https://github.com/JuliaLang/Compat.jl/pull/474 but wasn't able to:

The failure on Travis is because of something unrelated on master:
```
Error During Test at /home/travis/.julia/v0.7/Compat/test/runtests.jl:1079
  Test threw an exception of type MethodError
  Expression: Compat.IteratorSize(v) == Base.HasShape()
  MethodError: no method matching Base.HasShape()
  Stacktrace:
   [1] top-level scope at /home/travis/.julia/v0.7/Compat/test/runtests.jl:1079
```
Could this be merged so that I can fix https://github.com/JuliaIO/JSON.jl/pull/234 the way Steven would like?

TotalVerb:
Sure, performance can be fixed later.

stevengj:
@TotalVerb, I think @ScottPJones was asking for the Compat PR to be merged.

I really wish this had not been merged.  Now, instead of noisy deprecations, we have silent performance problems that we will have to remember to chase down later.

TotalVerb:
@stevengj I can revert if necessary, but I intend to revisit MemoryParserState anyway (as would be necessary to fix #232), so I figured it would not hurt that much. We can avoid tagging in the meantime.

ScottPJones:
I just came back to ask when the update to Compat would be tagged, so that I could push an update to this PR to use `codeunits` and `ncodeunits`.
As soon as that is tagged, I'll make another PR to JSON to use them (I don't like performance problems either, as you well know!).  However, regarding performance problems, there are a lot of much more significant ones in this code, such as creating strings just to calculate the length when there are `\u` escapes, which is why (in *this* PR), I didn't worry about fixing all of it's problems (having had keeping PRs focused rather beaten into me!)

src/Parser.jl
ScottPJones: Even using `codeunits` is not the best approach here.
Instead of having to create a string just to append it, it would be better to pass `b` to a `read_unicode_escape!(b, ps)` function, that directly `push!`ed the 1-4 bytes from the unicode escape.

ScottPJones: Another question - why do you think my change would force a copy?  It uses `Vector{UInt8}` on `v0.6`, which doesn't copy (and hence could be unsafe), and `codeunits` on `v0.7`, which doesn't copy either.
I think your objection was incorrect.

src/Parser.jl
ScottPJones: They don't make less copies either.  My change actually requires gets rid of the copies on v0.7, and doesn't add them for v0.6.

stevengj: Your version has makes more copies on 0.7 than 0.6, whereas mine does not.

stevengj:
Don't blame this on your tendency to raise unrelated issues (which are are doing *again*).   The problem with this PR has nothing to do with minimalism: it updates code which made no copies in 0.6 to code that makes copies in 0.7, rather than keeping the equivalent behavior.  It is a regression.  I'm not in favor of PRs that introduce regressions in order to silence deprecations, whereas avoiding the regression required little or no additional lines of code.

stevengj:
@TotalVerb, I'd prefer that you revert.

src/Parser.jl
ScottPJones: On v0.7, it calls `codeunits`.  Isn't that what your's does on v0.7?

ScottPJones:
"it updates code which made no copies in 0.6 to code that makes copies in 0.7, rather than keeping the equivalent behavior"
Stop making these incorrect statements, please.  This change was NOT a regression.
I made a change that did *not* introduce any additional copies, and was simpler than your proposed change.

stevengj:
Your change did introduce additional copies.  `Vector{UInt8}(codeunits(s))` makes a copy in 0.7.

PR #236, opened 2018-02-01
reduce number of string/buffer allocations

stevengj:
This eliminates a number of unnecessary string and buffer allocations.   It eliminates those from #234, and also eliminates several associated with Unicode escapes: rather than converting a `Char` to a `String` to get the UTF-8 representation, I write it directly to a buffer.

Most of the changes are deleting/simplifying code: my favorite kind of optimization :wink:.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.6
-Compat 0.45.0
+Compat 0.51.0
 Nullables 0.0.1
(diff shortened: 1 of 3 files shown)


stevengj:
The CI failure on 0.7 is an unrelated problem with the OffsetArrays test, which is not a regression (`master` currently has the same failure).

ScottPJones:
It looks like we are working on the same thing in parallel.
I'm testing out a fairly complete rewrite of all of the code in Parser.jl and specialize.jl, just haven't had time to make a PR out of it yet.

stevengj:
@TotalVerb, what do you think?  This PR reduces the code size and reduces the number of allocations, so it seems like a net win regardless of how much performance improvement it gives.

stevengj:
(Note that I don’t have commit access here.)

PR #237, opened 2018-02-02
JSONText(string) for embedding raw JSON in the output

stevengj:
This exports a `JSONText(s::String)` object that you can use to embed raw JSON in directly in the output.

For example, this would be useful in IJulia for JuliaLang/IJulia.jl#621

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,12 +71,18 @@ installed), you can pass `dicttype=DataStructures.OrderedDict` to
 maintain the insertion order of the items in the object.
 
 The `inttype` argument controls how integers are parsed.  If a number in a JSON
-file is recognized to be an integer, it is parsed as one; otherwise it is parsed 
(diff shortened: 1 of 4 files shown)


stevengj:
Unrelated CI failure on 0.7 (same as `master`) due to `OffsetArray` problem.

TotalVerb:
This will ignore the `indent=x` keyword argument for `json`, so I'm not 100% comfortable with it, as-is. Can we "enforce" the "no-extra whitespace" rule in the docstring (i.e. document that the JSON should be ensured to be valid and formatted correctly), and reparse and print for general `::SC` (only specializing on `::CompactContext`), so that users will not ever get unexpected results?

stevengj:
I updated the PR to document that valid JSON is required.  Also, as you suggest, I updated it to re-parse and prints for general contexts, only printing as-is for `CompactContext`.

PR #240, opened 2018-02-09
Fix all existing deprecation warnings

iamed2:
Special-cased Symbol for one of them, which felt icky but works.

### DIFF
--- a/src/Parser.jl
+++ b/src/Parser.jl
@@ -222,7 +222,7 @@ function parse_object(pc::ParserContext{DictType, <:Real}, ps::ParserState) wher
             # Read value
             value = parse_value(pc, ps)
             chomp_space!(ps)
-            obj[convert(keyT, key)] = value
+            obj[keyT === Symbol ? Symbol(key) : convert(keyT, key)] = value
             byteat(ps) == OBJECT_END && break
             skip!(ps, DELIMITER)
         end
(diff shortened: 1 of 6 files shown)


src/Parser.jl
```diff
@@ -222,7 +222,7 @@ function parse_object(pc::ParserContext{DictType, <:Real}, ps::ParserState) wher
             # Read value
             value = parse_value(pc, ps)
             chomp_space!(ps)
-            obj[convert(keyT, key)] = value
+            obj[keyT === Symbol ? Symbol(key) : convert(keyT, key)] = value
```
TotalVerb: This is really icky. Can we leave it and try to come up with a better solution later?

iamed2: What's the advantage to leaving it? This change at least stops everyone receiving a deprecation warning, and doesn't add any new behaviour or remove any existing behaviour.

TotalVerb: Presumably the deprecation of `String` to `Symbol` conversion was done for a reason, to discourage this kind of implicit conversion. We should rather provide an ability to specify an arbitrary function to process the key (along with the dict type).

ararslan: I think we should just do the easy adaptation to 0.7 now and refactor into something better later.

ararslan:
@TotalVerb Do you know what the CI failures are about here? Unrelated to these changes I assume, but as noted in https://github.com/JuliaLang/METADATA.jl/pull/13454#issuecomment-367239294, the package is failing its own tests now, on 0.6 and 0.7.

iamed2:
This line `String(take!(parse_string(ps, IOBuffer(maxsize=len))))` uses `maxsize` as a keyword but that's only available on 0.7. Needs to be Compatted.

This is what's currently breaking on 0.6.

PR #242, opened 2018-02-12
Rewrite parser code, fix number parsing bugs, improve tests, adapt to v0.7

ScottPJones:

PR #243, opened 2018-02-12
Use pointer instead of string / position to improve performance

ScottPJones:


stevengj:
I don't think switching to use raw pointers everywhere (as opposed to `codeunit(s, i)`, possibly with `@inbounds`) is a good idea.   Maybe localized in a central performance-critical function with benchmarks showing a clear impact from the switchover to pointers, but not as a wholesale replacement.

(In this case, it's not even possible to benchmark your code to see the impact of using raw pointers, since you've also changed lots of other things that will affect performance.)

ScottPJones:
To benchmark the code to see the impact of using raw pointers, all you have to do is compare this with #242.
I'd hoped that #242 would get merged first, and then this only would have the changes for using pointers, which did make a very large difference.
I disagree about using raw pointers everywhere, because  trying to use`codeunit(s, i)` means keeping a separate variable around, instead of just the pointer and an end pointer, which is part of why the performance is faster using pointers.

kmsquire:
Same comment as https://github.com/JuliaIO/JSON.jl/pull/242#issuecomment-493708459.

PR #246, opened 2018-02-22
Update for fieldnames change in v0.7

TotalVerb:
cc @ararslan 

The v0.6 test failure will need a Compat change, coming soon.


ararslan:
@TotalVerb What's the status on the Compat change? What in particular needs to be added?

TotalVerb:
I have a branch that's currently failing tests. It's the massive keyword argument

PR #247, opened 2018-02-24
Allow build failures on nightly in CI config

davidanthoff:


ararslan:
I'd prefer we not do this. #246 will fix 0.7 anyway.

PR #248, opened 2018-02-24
Handle IOBuffer signatures on 0.6 and 0.7

iamed2:
Tests should pass on 0.6 now.


TotalVerb:
see also https://github.com/JuliaLang/Compat.jl/pull/501

samoconnor:
It would be great if you could merge this and tag!
0.6 code is breaking in the wild: https://github.com/JuliaCloud/AWSCore.jl/issues/29

TotalVerb:
We can merge this and replace with the Compat change later.

PR #249, opened 2018-03-27
fixes for 0.7

rened:
Tests pass on 0.6 and 0.7 without warnings.

Needs https://github.com/JuliaCollections/DataStructures.jl/tree/aa/0.7

### DIFF
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -1,6 +1,8 @@
-__precompile__()
+# __precompile__() rene
 
 module JSON
+
+using Compat
 
 export json # returns a compact (or indented) JSON representation as a string
 export JSONText # string wrapper to insert raw JSON into JSON output
--- a/src/bytes.jl
+++ b/src/bytes.jl
@@ -44,7 +44,7 @@ const ESCAPES = Dict(
     LATIN_T      => TAB)
 
 const REVERSE_ESCAPES = Dict(reverse(p) for p in ESCAPES)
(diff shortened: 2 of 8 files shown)


src/JSON.jl
```diff
@@ -1,7 +1,9 @@
-__precompile__()
+# __precompile__() rene
```
stevengj: ?

rened: sorry about that, had commented it out for quicker development cycles.

src/bytes.jl
```diff
@@ -53,7 +53,11 @@ for c in 0x00:0xFF
     elseif haskey(REVERSE_ESCAPES, c)
         [BACKSLASH, REVERSE_ESCAPES[c]]
     elseif iscntrl(Char(c)) || !isprint(Char(c))
-        UInt8[BACKSLASH, LATIN_U, hex(c, 4)...]
+        if VERSION < v"0.7.0-DEV.4446"
+            UInt8[BACKSLASH, LATIN_U, hex(c, 4)...]
+        else
+            UInt8[BACKSLASH, LATIN_U, string(c, base=16, pad=4)...]
```
TotalVerb: can we Compat-ify this method of string? maybe not easily...

rened: Yes, looked into that, but I could not work around the stack overflows resulting from overwriting Base.string. I don't know if it'S possible to provide proper Compat versions of this.

I would suggest, as this is the only place where this is used, to merge as-is and revisit later.

PR #252, opened 2018-06-29
Fixes for v0.7

fredrikekre:
Most notably removes dependency on `Nullables` and `Compat`.


stevengj:
I would have preferred to keep 0.6 support until a little while after 0.7 is released, and then have a separate PR dropping compat...

fredrikekre:
Fair enough, should we then wait with the `Nullable{T}` -> `Union{T, Nothing}` change too?

KristofferC:
Last functional commit here was four months ago and it is always possible to backport something and release from another branch if the need arises.

Keno:
I agree, this package is fairly stable and if we're doing a somewhat breaking change in this package (`Nullable` -> `Union{T, Nothing}`), it would be good to get that in early, so other packages can adjust to these changes. Otherwise we'll have a second wave of instability as we upgrade packages. Releasing a hotfix from a branch should be just fine.

Keno:
@stevengj are you ok with that approach? Lots of packages depend on JSON.jl, so it would be good to get a compatible version out soon.

stevengj:
Okay.

PR #253, opened 2018-07-06
Allow any function to be used as a dicttype, so we can support DefaultDict

oxinabox:
DefaultDict has no 0 argument constructors.
Which means that to use it as the dicttype you really want to wrap it in any anon function.

This makes that legal.

I am not sure what `unparameterize_type` is meant to do.
So I am not sure if by bypassing it I have done evil


oxinabox:
Bump

stevengj:
LGTM; it shouldn't cause any regressions since it seems to be a strict relaxation of the type signatures.

TotalVerb:
Will require 0.7 at minimum, since we've already bumped REQUIRE. Do you need it tagged?

PR #255, opened 2018-08-07
nfields should be used on objects not types

KristofferC:
```
julia> struct SingletonType end

julia> a = SingletonType()
SingletonType()

julia> nfields(typeof(SingletonType()))
20
```


KristofferC:
Need new tags in test dependencies for this to pass.

Issue #256, opened 2018-08-30
lower missing to null?


YongHee-Kim:
I've noticed this behavior 
 ```julia
julia> JSON.json(nothing)
"null"

julia> JSON.json(missing)
ERROR: Cannot serialize type Missing
```
would lowering missing to "null" will make sence?

TotalVerb:
merged

PR #257, opened 2018-08-31
Add support for missing

YongHee-Kim:
https://github.com/JuliaIO/JSON.jl/issues/256

### DIFF
--- a/src/Writer.jl
+++ b/src/Writer.jl
@@ -263,6 +263,7 @@ function show_json(io::SC, s::CS, x::Union{Integer, AbstractFloat})
 end
 
 show_json(io::SC, ::CS, ::Nothing) = show_null(io)
+show_json(io::SC, ::CS, ::Missing) = show_null(io)
 
 function show_json(io::SC, s::CS, a::AbstractDict)
     begin_object(io)


KristofferC:
Isn't it a bit odd for two quite different Julia objects to serialize to the same representation?

YongHee-Kim:
Yes, I'm not knowledgeable enough to discuss this matter. 

Just thought this in user's perspective. `.json` is for storing data and we are encouraged to use Union{T, Missing} for handling data. So whenever someone wants to store something with missing, they will encounter errors. 
So if two different Julia objects to serialize to the same representation is problematic, maybe serialize `missing` to `null` and nothing to `"nothing"` would be a better solution?

quinnj:
I support this; JSON is limited in that it has a notion of `null`, but no way to distinguish the use in Julia of `nothing` as a software engineer's null or `missing` as missing data. I see JSON structures all the time that conflate the two. I think this is the most natural serialization strategy.

TotalVerb:
I think this is very reasonable. Could also consider a `lower(::Missing) = null` method, which might be cleaner.

TotalVerb:
As there are 4 votes in support, I think we should merge this.

YongHee-Kim:
Isn't this breaking change for 0.6? 
should add `using Missings` for 0.6

TotalVerb:
I haven't tagged this yet. Since tests aren't being run on 0.6, we should
just drop support.

TotalVerb:
Actually, 0.6 is already not present in REQUIRE, so there is no action required (unless we want to backport a fix to 0.6?).

PR #261, opened 2018-10-26
Fix integer parsing with number-accepted characters that are not digits

### LINKED ISSUE #260: parsed invalid string rather than throwing an error
By providing an incorrect command line argument I provided `JSON.parse` with a date format string `yyyy-mm-dd` which I believe should have been invalid.  The parsing for some reason seems to return an integer.

For example
```julia
JSON.parse("1997-03-03")
>> 2022328303
```

TotalVerb:
Fixes #260.

### DIFF
--- a/src/Parser.jl
+++ b/src/Parser.jl
@@ -128,7 +128,7 @@ end
 
 
 # Throws an error message with an indicator to the source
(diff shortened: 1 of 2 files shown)


TotalVerb:
Absent objections, I'll merge and tag in 24 hours.

PR #264, opened 2018-11-04
Improve number parsing: use a PushVector to work around slow push! in Base

KristofferC:
Long term it might be better to accumulate numbers on the fly without storing them in an intermediate Vector. In the meantime, this approximately doubles the performance of parsing numbers. Using the benchmark

```jl
using JSON
using BenchmarkTools

a = string("[", join([rand(Int) for i in 1:10^3], ","), "]");
@btime JSON.parse($a)
```

we have before:

```jl
julia> @btime JSON.parse($a)
  228.841 μs (1016 allocations: 32.33 KiB)
```

and after:

```jl
julia> @btime JSON.parse($a)
  119.424 μs (1013 allocations: 32.25 KiB)
```


TotalVerb:
Does this still give improvement after #265?

KristofferC:
Yes, but only for the `StreamingParser` case (since after #265, the `MemoryParser` doesn't use this at all).

PR #265, opened 2018-11-04
restore special case parsing of numbers for in memory JSON

KristofferC:
I was trying to figure out why the `from` and `to` variables where threaded through the number routines since it seemed like they were always `1` and `length(ps.utf8array)` respectively. Turns out there is a currently unused special cased number parser is being failed to dispatched to after the introduction of `ParserContext` (https://github.com/JuliaIO/JSON.jl/pull/224 cc @kmsquire).

This restores the special case for a nice performance boost:

```jl
using JSON
using BenchmarkTools

a = string("[", join([rand(Int) for i in 1:10^3], ","), "]");
@btime JSON.parse($a)
```

Before:

```
julia> @btime JSON.parse($a)
  246.210 μs (1016 allocations: 32.33 KiB)
```

After:

```
julia> @btime JSON.parse($a)
  89.930 μs (1011 allocations: 32.11 KiB)
```

### DIFF
--- a/src/Parser.jl
+++ b/src/Parser.jl
@@ -20,13 +20,12 @@ abstract type ParserState end
 mutable struct MemoryParserState <: ParserState
     utf8::String
     s::Int
-    utf8array::Vector{UInt8}
 end
 
 # it is convenient to access MemoryParserState like a Vector{UInt8} to avoid copies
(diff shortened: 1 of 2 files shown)


TotalVerb:
LGTM. Not sure why the codecov is failing, since this improves coverage.

KristofferC:
Coverage is fubar...

Issue #169, opened 2016-09-10
Reader generalization


TotalVerb:
After #165, to implement #168 properly still requires reader support. A similar analogous as done for writers can be done for readers also. The big roadblocks here are:

- unlike for writers, there's no unified `Base` parsing/reading API; https://github.com/JuliaLang/julia/pull/16981 seems to start on this
- for writers, the monadic-style composition is clear; for readers, this is less clearcut

klickverbot:
Any progress on this in the meantime? I need to handle a Python-generated file that contains both `NaN` as a number (violating the JSON spec) and "NaN" as a string. Currently, I'm using Python's parser via PyCall to get around this, but obviously I'd rather like to avoid that.

PR #280, opened 2019-05-13
RFC: Add `allownan` keyword argument to parse() (Fixes #168)

### LINKED ISSUE #168: Reading NaN and Inf
We've got some external non-julia code producing JSON, with some Float64 values which happen to be NaN.  We can produce null for these, but it's unfortunate that they then can't be distinguished from Inf.

This has been discussed before - see #75 and https://github.com/JuliaLang/JSON.jl/pull/56#issuecomment-39531579 but it would be nice to have the option for strict vs flexible handling of NaN and Inf (for example the way that the latest RapidJson does this is to [use a writer flag](https://github.com/miloyip/rapidjson/blob/328ead0e173a960e0fc164e102cf8e1734c16380/include/rapidjson/writer.h#L66))

Is there any interest in supporting this kind of thing?  Would a PR be likely to be accepted?

CC @janzill

kmsquire:
* Allows reading out-of-spec JSON files which include `Nan`, `Infinity`, and `-Infinity`

The ideal solution to reading out-of-spec JSON files was proposed by @TotalVerb in #169.

Until that is implemented, this provides a pragmatic solution for people to read these types of JSON files when they don't have control over the file generation.

### DIFF
--- a/src/Parser.jl
+++ b/src/Parser.jl
@@ -37,7 +37,7 @@ mutable struct StreamingParserState{T <: IO} <: ParserState
 end
 StreamingParserState(io::IO) = StreamingParserState(io, 0x00, true, PushVector{UInt8}())
 
-struct ParserContext{DictType, IntType} end
+struct ParserContext{DictType, IntType, AllowNanInf} end
 
 """
 Return the byte at the current position of the `ParserState`. If there is no
@@ -169,11 +169,11 @@ function parse_value(pc::ParserContext, ps::ParserState)
     elseif byte == ARRAY_BEGIN
         parse_array(pc, ps)
     else
-        parse_jsconstant(ps::ParserState)
+        parse_jsconstant(pc, ps)
     end
 end
 
-function parse_jsconstant(ps::ParserState)
+function parse_jsconstant(::ParserContext{<:Any,<:Any,AllowNanInf}, ps::ParserState) where AllowNanInf
     c = advance!(ps)
     if c == LATIN_T      # true
         skip!(ps, LATIN_R, LATIN_U, LATIN_E)
@@ -184,6 +184,12 @@ function parse_jsconstant(ps::ParserState)
     elseif c == LATIN_N  # null
         skip!(ps, LATIN_U, LATIN_L, LATIN_L)
         nothing
+    elseif AllowNanInf && c == LATIN_UPPER_N
+        skip!(ps, LATIN_A, LATIN_UPPER_N)
+        NaN
+    elseif AllowNanInf && c == LATIN_UPPER_I
+        skip!(ps, LATIN_N, LATIN_F, LATIN_I, LATIN_N, LATIN_I, LATIN_T, LATIN_Y)
+        Inf
     else
         _error(E_UNEXPECTED_CHAR, ps)
     end
@@ -207,7 +213,7 @@ function parse_array(pc::ParserContext, ps::ParserState)
 end
 
 
-function parse_object(pc::ParserContext{DictType, <:Real}, ps::ParserState) where DictType
+function parse_object(pc::ParserContext{DictType,<:Real,<:Any}, ps::ParserState) where DictType
     obj = DictType()
     keyT = keytype(typeof(obj))
 
@@ -323,7 +329,7 @@ end
 Parse an integer from the given bytes vector, starting at `from` and ending at
 the byte before `to`. Bytes enclosed should all be ASCII characters.
 """
-function int_from_bytes(pc::ParserContext{<:Any,IntType},
+function int_from_bytes(pc::ParserContext{<:Any,IntType,<:Any},
                         ps::ParserState,
                         bytes,
                         from::Int,
@@ -364,11 +370,21 @@ function number_from_bytes(pc::ParserContext,
 end
 
 
-function parse_number(pc::ParserContext, ps::ParserState)
+function parse_number(pc::ParserContext{<:Any,<:Any,AllowNanInf}, ps::ParserState) where AllowNanInf
     # Determine the end of the floating point by skipping past ASCII values
     # 0-9, +, -, e, E, and .
     number = ps.utf8array
     isint = true
+    negative = false
+
+    c = current(ps)
+
+    # Parse and keep track of initial minus sign (for parsing -Infinity)
+    if AllowNanInf && c == MINUS_SIGN
+        push!(number, UInt8(c)) # save in case the next character is a number
+        negative = true
+        incr!(ps)
+    end
 
     @inbounds while hasmore(ps)
         c = current(ps)
@@ -378,6 +394,10 @@ function parse_number(pc::ParserContext, ps::ParserState)
         elseif c in (PLUS_SIGN, LATIN_E, LATIN_UPPER_E, DECIMAL_POINT)
             push!(number, UInt8(c))
             isint = false
+        elseif AllowNanInf && c == LATIN_UPPER_I
+            infinity = parse_jsconstant(pc, ps)
+            resize!(number, 0)
+            return (negative ? -infinity : infinity)
         else
             break
         end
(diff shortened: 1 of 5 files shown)


kmsquire:
There are no tests yet.  I'll wait for any discussion on whether or not this is desired first.

ararslan:
Seems okay to me in principle but I'm all that familiar with the JSON spec nor the internals of this package so I'm probably not the best person to review this.

TotalVerb:
I think that if we have a permissive mode like this we should also support the Python output nan, inf, -inf (note case). I'm not sure if any Python code generates this output though.

kmsquire:
> I think that if we have a permissive mode like this we should also support the Python output nan, inf, -inf (note case). I'm not sure if any Python code generates this output though.

According to the [Python `json` package docs](https://docs.python.org/3/library/json.html#basic-usage), it outputs the JavaScript values:

> If `allow_nan` is false (default: `True`), then it will be a `ValueError` to serialize out of range `float` values (`nan`, `inf`, `-inf`) in strict compliance of the JSON specification. If `allow_nan` is true, their JavaScript equivalents (`NaN`, `Infinity`, `-Infinity`) will be used.

[simplejson](https://simplejson.readthedocs.io/en/latest/) and [rapidjson](https://python-rapidjson.readthedocs.io/en/latest/dumps.html?highlight=nan) follow the same convention.  `ujson` doesn't output any of these values.

Google's JavaScript GSON module also accepts `NaN`, `Infinity`, `-Infinity`.  It does not output them by default, but allows this behavior to be overridden (see 
https://static.javadoc.io/com.google.code.gson/gson/2.6/com/google/gson/GsonBuilder.html#serializeSpecialFloatingPointValues()).

Given this, I don't think we need to support standard Python values in JSON.

kmsquire:
While looking through some of the packages in the previous comment, I ran across this snippet (from https://tools.ietf.org/html/rfc7159.html#section-9):

```
   A JSON parser transforms a JSON text into another representation.  A
   JSON parser MUST accept all texts that conform to the JSON grammar.
   A JSON parser MAY accept non-JSON forms or extensions.
```

Given this, I think what we're doing here is not too unreasonable.

That said
* I still support @TotalVerb's original idea to make this a configurable.  I just don't have time to work on that.
* An alternative would be to fork this package and create a `JSON5` package (see https://json5.org/), which accepts `NaN`, `Infinity`, and a number of other JSON extensions (comments, trailing commas, keys without quotes, single quoted strings, etc.).  Again, takes time.  I don't think json5 is widely used yet.

kmsquire:
I'll add tests, and see if anyone else wants to review.

kmsquire:
1. I added tests
2. I refactored the code to reduce code duplication.  @TotalVerb, can you review again if you get the chance?

The codecov results are a bit hard to understand--it looks like we're not providing the right line numbers, and/or some of the code that was added was simply inlined later.

src/specialized.jl
```diff
@@ -120,19 +120,32 @@ function parse_string(ps::MemoryParserState, b::IOBuffer)
...
+    @inbounds c = ps[p]
+
+    # Parse and keep track of initial minus sign (for parsing -Infinity)
+    if AllowNanInf && c == MINUS_SIGN
+        negative = true
+        p += 1
+    end
 
     # Determine the end of the floating point by skipping past ASCII values
-    # 0-9, +, -, e, E, and .
```
TotalVerb: why is this removed in comment? - can occur after the initial position, e.g. in 2e-4

test/parser/nan-inf.jl
```diff
@@ -0,0 +1,14 @@
+@testset begin
+    test_str = """{"x": NaN, "y": Infinity, "z": -Infinity}"""
+    @test_throws ErrorException JSON.parse(test_str)
+    val = JSON.parse(test_str, allownan=true)
+    @test val["x"] === NaN
```
TotalVerb: I'm a little bugged about the === vs == here (also, not all NaNs are ===). Can we use `isnan` or `isequal` consistently?

```
julia> reinterpret(Float64, 9221120237041090561)
NaN

julia> reinterpret(Float64, 9221120237041090561) === NaN
false
```

src/specialized.jl
```diff
@@ -120,19 +120,32 @@ function parse_string(ps::MemoryParserState, b::IOBuffer)
...
     e = length(ps)
     isint = true
+    negative = false
+
+    @inbounds c = ps[p]
+
+    # Parse and keep track of initial minus sign (for parsing -Infinity)
+    if AllowNanInf && c == MINUS_SIGN
+        negative = true
+        p += 1
```
TotalVerb: I can't test this right now but I suspect that this breaks parsing regular negative floats in AllowNanInf mode. I might be wrong but the code seems to read like it skips the minus sign here, even if not followed by I. Could you add a test for that?

src/Parser.jl
```diff
@@ -364,11 +370,21 @@ function number_from_bytes(pc::ParserContext,
...
     isint = true
+    negative = false
+
+    c = current(ps)
+
+    # Parse and keep track of initial minus sign (for parsing -Infinity)
+    if AllowNanInf && c == MINUS_SIGN
+        push!(number, UInt8(c)) # save in case the next character is a number
+        negative = true
+        incr!(ps)
```
TotalVerb: same comment here as below; does this still preserve correct behaviour with regular negative numbers?

src/Parser.jl
```diff
@@ -421,20 +442,22 @@ end
...
+               allownan=false)
+    pc = _get_parsercontext(dicttype, inttype, allownan)
     ps = StreamingParserState(io)
     parse_value(pc, ps)
 end
 
 function parsefile(filename::AbstractString;
                    dicttype=Dict{String, Any},
                    inttype::Type{<:Real}=Int64,
+                   allownan=false,
```
stevengj: Maybe we should default to `allownan=true`, i.e. permissive by default?

test/parser/nan-inf.jl
kmsquire: > I'm a little bugged about the `===` vs `==` here (also, not all `NaN`s are `===`). Can we use `isnan` or `isequal` consistently?

Sure.  I had forgotten that `isequal` compared treated `NaN`s as equal.

src/Parser.jl
kmsquire: This one had issues.  Parsing negative numbers is okay, but I forgot to reset the parse buffer, so the `-` sign was still in the buffer at the start of the next parse.

The current test catches this, but I'll add a more specific one.

src/Parser.jl
kmsquire: I was actually going to suggest this as well.  There's precedent--e.g., the Google GSON library does this, even though it doesn't produce these values by default (and neither do we).

src/specialized.jl
kmsquire: This one was actually ok (`s` above marks the start of the number).

src/specialized.jl
kmsquire: Fixed in latest push.  (I had remembered this, but forgot to revert this change.)

test/parser/nan-inf.jl
kmsquire: Using `isequal` for all tests in the latest commit.

src/specialized.jl
kmsquire: (Not broken, but tests added.)

src/Parser.jl
kmsquire: Tests have been updated.

src/Parser.jl
kmsquire: The last commit sets `allownan=true` by default.  I think this is a good change, but want to make sure there's consensus before merging.

src/Parser.jl
```diff
@@ -323,7 +329,7 @@ end
 Parse an integer from the given bytes vector, starting at `from` and ending at
 the byte before `to`. Bytes enclosed should all be ASCII characters.
 """
-function int_from_bytes(pc::ParserContext{<:Any,IntType},
+function int_from_bytes(pc::ParserContext{<:Any,IntType,<:Any},
```
kmsquire: I don't think this change (here or elsewhere) is strictly necessary (unspecified type parameters should default to `<:Any`), but it shouldn't hurt either.

kmsquire:
Okay, thanks @TotalVerb.  Merging!

PR #288, opened 2019-07-02
Made `show_json` for composite types more flexible via `propertynames`/`getproperty`.

twadleigh:
This lets `show_json(..., ::CompositeTypeWrapper)` potentially do more interesting things by providing a custom `getproperty` method.

~It also has `show_json` apply `something` to values of type `Some`.~ I decided it would be safer not to assume this is always the desired behavior.

### DIFF
--- a/src/Writer.jl
+++ b/src/Writer.jl
@@ -20,7 +20,7 @@ struct CompositeTypeWrapper{T}
 end
 
 CompositeTypeWrapper(x, syms) = CompositeTypeWrapper(x, collect(syms))
-CompositeTypeWrapper(x) = CompositeTypeWrapper(x, fieldnames(typeof(x)))
+CompositeTypeWrapper(x) = CompositeTypeWrapper(x, propertynames(x))
 
 """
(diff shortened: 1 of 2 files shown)


src/Writer.jl
```diff
@@ -282,7 +282,7 @@ end
 function show_json(io::SC, s::CS, x::CompositeTypeWrapper)
     begin_object(io)
     for fn in x.fns
-        show_pair(io, s, fn, getfield(x.wrapped, fn))
+        show_pair(io, s, fn, getproperty(x.wrapped, fn))
```
KristofferC: Just `x.wrapped.fn`?

twadleigh: That wouldn't work here. `fn` is a variable whose value is a `Symbol` that is the property name. It is not itself the property name.

KristofferC: Oh sorry, I missed the context.

TotalVerb:
LGTM. Thanks!

PR #290, opened 2019-07-16
Restructure Project.toml, bump version for next release

ararslan:


fredrikekre:
Should probably have been 0.20.1 but 🤷‍♂

ararslan:
Yeah probably. I did minor because features were added since 0.20.0, even though it's technically okay to do otherwise given that the version is < 1.0.

KristofferC:
> even though it's technically okay to do otherwise given that the version is < 1.0.

Well, it is more than technically okay. It is that all code using semver for compat now has to check if they are compatible with the new (according to the maintainers) breaking version and tag a new version to use it. Which is a lot of unnecessary work if there wasn't any breaking change.

ararslan:
Okay, I'll remember that for next time.

Issue #298, opened 2020-01-05
Document custom serialization


travigd:
There's not very much documentation around the custom serialization API and it took a bit of hunting around in the code to figure out how to make something work.

```julia
using JSON

struct MySerialization <: JSON.CommonSerialization end

# Define specific method for Missing
JSON.show_json(io::JSON.StructuralContext, s::MySerialization, x::Missing) =
    JSON.show_json(io, s, "~missing~")

myjson(x) = sprint(JSON.show_json, MySerialization(), x)
println(myjson(Dict("foo" => missing)))
# {"foo":"~missing~"}
```

PR #300, opened 2020-01-06
Add documentation for StructuralContext to README

TotalVerb:
Partially address #298 (still need docs for `Serialization`, but I only had time to write these so far).

I apologize for the lack of documentation when I implemented these features — it had been on my TODO list but slipped from the mind. I will make another PR with docs and examples for `Serialization` soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,3 +106,167 @@ JSON.lower(p::Point2D) = [p.x, p.y]
 
 Define a custom serialization rule for a particular data type. Must return a
 value that can be directly serialized; see help for more details.
+
+### Customizing JSON
+
+Users may find the default behaviour of JSON inappropriate for their use case. In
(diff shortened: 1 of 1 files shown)


travigd:
This looks awesome! Thank you for doing this!

It might be worth investing in setting up _proper_ docs (à la Documenter.jl) because it might make the readme a bit bloated (definitely something that could be punted down the road though).

TotalVerb:
Indeed, I agree that proper docs would be the right solution! In general, there is a lot of work that could be done to modernize this package (most of it was written back in the Julia 0.3/0.4 days), and adhering to new conventions is one thing we have been lagging on. But I guess it is better to have the documentation somewhere first.

We will keep #298 open until we have docs for `Serialization` too, which is perhaps actually even more important.

PR #305, opened 2020-04-15
Print longer arrays without newlines

### LINKED ISSUE #291: Dense printing for long arrays
It would be nice to be able to keep the general pretty formatting, but remove newlines in long arrays. 

Say, if the array is longer than length `densearraythreshold` (which could default to 10), don't print with newlines

IanButterworth:
Saving JSON files with long arrays results in very long documents currently because of all the newlines.
This only prints newlines after array elements if the length of the array is > 5.

i.e.

```
{
  "x": [1,2,3,4,5,6,7,8,9,10]
}
```
instead of
```
{
  "x": [
    1,
    2,
    3,
    4,
    5,
    6,
    7,
    8,
    9,
    10
 ]
}
```

Fixes #291

### DIFF
--- a/src/Writer.jl
+++ b/src/Writer.jl
@@ -129,18 +129,18 @@ write(io::StringContext, char::Char) =
 =#
 
 """
-    indent(io::StructuralContext; newline = true)
+    indent(io::StructuralContext)
 
 If appropriate, write a newline to the given context, then indent it by the
 appropriate number of spaces. Otherwise, do nothing.
 """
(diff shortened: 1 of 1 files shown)


ianshmean:
The travis error is just a Pkg download glitch

ianshmean:
This shouldn't be merged as-is. Needs rethinking. 
Also JSON3.jl doesn't print newlines for arrays, so I'm giving that a go for

quinnj:
Closing as stale and could use a more modern approach to how to configure

PR #310, opened 2020-09-07
revert 1.0 bump and tag + release 0.21.1 instead

KristofferC:
This allows us to get a new JSON version out to everyone with the latest bugfixes and improvements.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "JSON"
 uuid = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
-version = "1.0.0"
+version = "0.21.1"
 
 [deps]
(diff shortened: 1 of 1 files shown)


KristofferC:
Unless someone objects I will merge and tag a new version in not too long.

TotalVerb:
This will just move the maintenance burden to the 1.0 "branch" instead of the 0.21 "branch". I think this LGTM to get the fix out, but unfortunately we'll need to continue tagging each fix with both version numbers, it seems.

KristofferC:
There is no 1.0 branch anymore though, since 1.0 was never released.

TotalVerb:
Ah ok, then LGTM. Let's hold off on the 1.0 tag then.

PR #321, opened 2021-05-07
Add eof check for streaming parser

### LINKED ISSUE #314: Unexpected end of input error when parsing input streams
When parsing JSON values from an input stream like this

```julia
using JSON

while !eof(stdin)
  JSON.parse(stdin)
end
```
depending on the type of a value in the input, parsing either succeeds or fails.

For example, if the code above is in `parse.jl`, and file `1.in` contains just `5` (no spaces before or after the number), this works:

    $ julia parse.jl < 1.in

But if `2.in` contains `[5]` instead (no spaces before or after the list), then there is an error:

    $ julia parse.jl < 2.in
    ERROR: LoadError: Unexpected end of input
     ...when parsing byte with value '10'

wardlem:
## Summary

Adds an end of file check to the streaming parser.

## Motivation

```julia
# This works
JSON.parse("123")

# This also wrks
JSON.parse(IOBuffer("123\n"))

# This breaks
JSON.parse(IOBuffer("123"))
```

Fixes #314

### DIFF
--- a/test/regression/issue314.jl
+++ b/test/regression/issue314.jl
@@ -1,0 +1,2 @@
+@test JSON.parse(IOBuffer("123")) == 123
+@test JSON.parse(IOBuffer("1.5")) == 1.5
--- a/src/Parser.jl
+++ b/src/Parser.jl
@@ -105,7 +105,7 @@ Return `true` if there is a current byte, and `false` if all bytes have been
 exausted.
 """
 @inline hasmore(ps::MemoryParserState) = ps.s ≤ length(ps)
-@inline hasmore(ps::StreamingParserState) = true  # no more now ≠ no more ever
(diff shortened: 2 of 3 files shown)


test/regression/issue314.jl
```diff
@@ -0,0 +1,2 @@
+@test JSON.parse(IOBuffer("123")) == 123
```
KristofferC, line 1: No need to create a new file for this, just add it to e.g. https://github.com/JuliaIO/JSON.jl/blob/1c24980c320842e17720c173be767c161f36feb9/test/runtests.jl#L36

pitipatw:
Is this issue resolved? 
I also got an error that seems to be related to this one: 

  Unexpected end of input
│     ...when parsing byte with value '0'

PR #327, opened 2021-07-30
Update Parsers support

quinnj:
Update the Parsers compat versions to `"1, 2"`. As of v1.0.5,
`Parsers.tryparse` ensured the full input buffer was consumed when
parsing a value (as mentioned desirable in the now-deleted comment in
`float_from_bytes`). So dropping support for older Parsers versions
allows us to trust `Parsers.tryparse`, and those pre-1.0 versions are
pretty old at this point anyway.

I locally ran the JSON.jl tests under both Parsers v2.0.1 and v1.1.1 and
ensured both were passing, so it shouldn't matter which version of
Parsers ends up getting installed if there are other packages which
don't 2.0 yet.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -18,7 +18,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
 julia = "0.7, 1.0"
-Parsers = "0.1, 0.2, 0.3, 1"
+Parsers = "1, 2"
 
 [targets]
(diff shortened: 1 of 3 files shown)


KristofferC:
> I locally ran the JSON.jl tests under both Parsers v2.0.1 and v1.1.1 and
ensured both were passing, 

CI seems to fail so perhaps there is some difference with your local state and CI?

KristofferC:
Ah, I see https://github.com/JuliaData/Parsers.jl/pull/85 now.

quinnj:
Ok, this looks good now.

KristofferC (approved):

PR #331, opened 2021-11-19
Add singleton support

Wynand:
Currently singletons don't work with `lower`, and this adds support for that

Effectively the singletons are printed as their class with "()" as a suffix to show that they're an instance of the singleton instead of the singleton type


Wynand:
This issue may be relevant: https://github.com/JuliaIO/JSON.jl/issues/203

PR #333, opened 2021-11-22
See alternative package JSON3

PallHaraldsson:
[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # JSON.jl
 
-This package provides for parsing and printing JSON in pure Julia.
+This package provides for parsing and printing JSON in pure Julia. See also alternative package [JSON3.jl](https://github.com/quinnj/JSON3.jl).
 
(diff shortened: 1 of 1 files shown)


PallHaraldsson:
I believe JSON3 is considered stable. And probably this package, while not yet 1.0. I know Dash.jl uses both (just dropped JSON2). I think that may be a historical accident, and all could and should use JSON3 now?

README.md
```diff
@@ -1,6 +1,6 @@
 # JSON.jl
 
-This package provides for parsing and printing JSON in pure Julia.
+This package provides for parsing and printing JSON in pure Julia. See also alternative packages [JSON3.jl](https://github.com/quinnj/JSON3.jl), and [Serde.jl](https://github.com/bhftbootcamp/Serde.jl) (based on the Serde Rust library that handles more formats than just JSON).
```
giordano, line 3: The claim that Serde.jl is based on the rust library is blatantly false and misleading. In fact, Serde.jl uses this package.

PR #336, opened 2022-01-25
Update doc with correct parameterized function syntax

o314:
This format is not supported since a long time.
Should we remap Associative to AbstractDict too ?

### DIFF
--- a/src/Parser.jl
+++ b/src/Parser.jl
@@ -425,11 +425,11 @@ function _get_parsercontext(dicttype, inttype, allownan, null)
 end
 
 """
-    parse{T<:Associative}(str::AbstractString;
+    parse(str::AbstractString;
                           dicttype::Type{T}=Dict,
                           inttype::Type{<:Real}=Int64,
                           allownan::Bool=true,
-                          null=nothing)
+                          null=nothing) where {T<:Associative}
 
 Parses the given JSON string into corresponding Julia types.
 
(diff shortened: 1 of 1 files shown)


singularitti:
Hi, I guess you should also align the first line like `parse(str::AbstractString;` with the rest of the lines.

o314:
This should be done now.
*Nota that mentions to Associative have not been reviewed*

src/Parser.jl
```diff
@@ -425,11 +425,11 @@ function _get_parsercontext(dicttype, inttype, allownan, null)
...
-    parse{T<:Associative}(str::AbstractString;
-                          dicttype::Type{T}=Dict,
-                          inttype::Type{<:Real}=Int64,
-                          allownan::Bool=true,
-                          null=nothing)
+    parse(str::AbstractString;
+          dicttype::Type{T}=Dict,
+          inttype::Type{<:Real}=Int64,
+          allownan::Bool=true,
+          null=nothing) where {T<:Associative}
```
fredrikekre, line 432:
```suggestion
          null=nothing) where {T<:AbstractDict}
```

src/Parser.jl
```diff
@@ -456,11 +456,11 @@ function parse(str::AbstractString;
...
-    parse{T<:Associative}(io::IO;
-                          dicttype::Type{T}=Dict,
-                          inttype::Type{<:Real}=Int64,
-                          allownan=true,
-                          null=nothing)
+    parse(io::IO;
+          dicttype::Type{T}=Dict,
+          inttype::Type{<:Real}=Int64,
+          allownan=true,
+          null=nothing) where {T<:Associative}
```
fredrikekre, line 463:
```suggestion
          null=nothing) where {T<:AbstractDict}
```

PR #342, opened 2022-04-28
update README with IO example

uhtewihptihwue:
As suggested by the activity on this gist https://gist.github.com/silgon/0ba43e00e0749cdf4f8d244e67cd9d6a, json file IO is a frequently used basic feature of this library. I believe that making such information available up front would be valuable for new users of this library


quinnj:
The README now includes an example like the one proposed here

PR #345, opened 2022-05-16
Introduce a recursive cycle check when writing

quinnj:
I noticed that writing can blow up when there are recusrive objects that reference each other.
(For example, in HTTP.jl, `Response` and `Request` reference each other)
This PR proposes a simple API for the `CompactContext` and `PrettyContext` where
`objectid` of objects will be tracked recursively when writing and it's configurable
what should be written out when a recursive cycle is detected.

Custom contexts can "hook in" to this behavior by subtyping `RecursiveCheckContext` and
including the required fields (see docs for new context). Otherwise, there shouldn't
be any functional change to APIs in any way.

cc: @vilterp (oops, wrong ping the 1st time)

### DIFF
--- a/src/Writer.jl
+++ b/src/Writer.jl
@@ -83,27 +83,43 @@ abstract type JSONContext <: StructuralContext end
 """
 Internal implementation detail.
 
+To handle recursive references in objects/arrays when writing, by default we want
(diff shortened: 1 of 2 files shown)


quinnj:
@KristofferC or @fredrikekre, could I bug you to review here?

PR #359, opened 2024-01-04
Fix DEFAULT_PARSERCONTEXT, speed up parsing.

Zentrik:
Unfortunately this was never updated in #280 to be the correct default, and so has been wrong ever since introduction.

### DIFF
--- a/src/Parser.jl
+++ b/src/Parser.jl
@@ -415,9 +415,9 @@ function unparameterize_type(T::Type)
 end
 
 # Workaround for slow dynamic dispatch for creating objects
-const DEFAULT_PARSERCONTEXT = ParserContext{Dict{String, Any}, Int64, false, nothing}()
+const DEFAULT_PARSERCONTEXT = ParserContext{Dict{String, Any}, Int64, true, nothing}()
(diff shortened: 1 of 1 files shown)


src/Parser.jl
```diff
@@ -415,9 +415,9 @@ function unparameterize_type(T::Type)
 end
 
 # Workaround for slow dynamic dispatch for creating objects
-const DEFAULT_PARSERCONTEXT = ParserContext{Dict{String, Any}, Int64, false, nothing}()
+const DEFAULT_PARSERCONTEXT = ParserContext{Dict{String, Any}, Int64, true, nothing}()
```
sjkelly, line 418: If we change the default to be a non-strict spec, we will certainty want to update the docs and make a version bump. The Python docs may serve as a guide for communicating this: https://github.com/JuliaIO/JSON.jl/pull/280#issuecomment-492385462

Zentrik: Unless I'm missing something, this is a default in the sense that there is a fast path for it. This change doesn't change the result but is surely a performance improvement. If you look further down you'll see that these are the default parameters for `parse`.

Zentrik:
Bump

PR #368, opened 2024-11-14
Remove false statement about Serde.jl

giordano:
Ref: https://github.com/JuliaIO/JSON.jl/pull/333#discussion_r1841396603.


PallHaraldsson:
Thanks, but leave in "more formats" part? I'm not sure where I can access the suggest edit feature...

PR #381, opened 2025-10-03
Add `write_json` to improve compatibility

nhz2:
This PR adds a `JSON.write_json` function with the same functionality and keyword arguments as the new `JSON.json` function. This function is provided for backward compatibility when upgrading from 
older versions of JSON.jl where the `json` function signature differed.

The idea is that `JSON.write_json` will be backported to v0.21.5

EDIT: Changed the name from `writefile` to `write_json`

### DIFF
--- a/src/JSON.jl
+++ b/src/JSON.jl
@@ -104,6 +104,67 @@ print(io::IO, obj, indent=nothing) = json(io, obj; pretty=something(indent, 0))
 print(a, indent=nothing) = print(stdout, a, indent)
 @doc (@doc json) print
 
+"""
+    JSON.writefile(Vector{UInt8}, x; kw...)::Vector{UInt8}
+    JSON.writefile(io::IO, x; kw...)
+    JSON.writefile(filename::AbstractString, x; kw...)
+
+Serialize `x` to JSON format. This function provides the same functionality as [`JSON.json`](@ref)
+
+The first method returns the JSON output as a `Vector{UInt8}`.
+The second method writes JSON output to an `IO` object.
+The third method writes JSON output to a file specified by `filename` which will
+be created if it does not exist yet or overwritten if it does exist.
+
+This function is provided for backward compatibility when upgrading from 
+older versions of JSON.jl where the `json` function signature differed.
+For new code, prefer using [`JSON.json`](@ref) directly.
+
+All methods accept the same keyword arguments as [`JSON.json`](@ref):
+
+- `omit_null`, `omit_empty`, `allownan`, `jsonlines`, `pretty`, `inline_limit`
+- `ninf`, `inf`, `nan`, `float_style`, `float_precision`, `bufsize`, `style`
+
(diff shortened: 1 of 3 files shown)


src/JSON.jl
```diff
@@ -104,6 +104,67 @@ print(io::IO, obj, indent=nothing) = json(io, obj; pretty=something(indent, 0))
...
+JSON.writefile("output.json", [1, 2, 3])
+
+# Get as bytes
+bytes = JSON.writefile(Vector{UInt8}, Dict("hello" => "world"))
+String(bytes)  # "{\"hello\":\"world\"}"
+```
+"""
+function writefile end
+
+function writefile(::Type{Vector{UInt8}}, x; pretty::Union{Integer,Bool}=false, kw...)
```
quinnj, line 147: I'm not following why we want this or the 2nd definitions? `writefile` that doesn't write to a file and just returns a byte vector?

nhz2: I want to avoid extra conversions between `String` and `Vector{UInt8}` for use with Zarr.jl and SmallZarrGroups.jl

I'm also trying to match the style of the `writefile` function in Parquet2.jl https://expandingman.gitlab.io/Parquet2.jl/#Writing-Data

nhz2: Here is a benchmark on the cost of doing the extra String to Vector{UInt8} conversion:

```julia-repl
julia> using Chairmarks, JSON, Random

julia> @be randstring(10000) Vector{UInt8}(JSON.json(_)) seconds=2
Benchmark: 52143 samples with 4 evaluations
 min    4.807 μs (7 allocs: 19.852 KiB)
 median 4.874 μs (7 allocs: 19.852 KiB)
 mean   5.566 μs (7 allocs: 19.856 KiB, 0.82% gc time)
 max    15.658 ms (7 allocs: 19.883 KiB, 99.92% gc time)

julia> @be randstring(10000) JSON.write_json(Vector{UInt8}, _) seconds=2
Benchmark: 50390 samples with 7 evaluations
 min    2.971 μs (4 allocs: 9.914 KiB)
 median 3.006 μs (4 allocs: 9.914 KiB)
 mean   3.394 μs (4 allocs: 9.919 KiB, 0.82% gc time)
 max    8.692 ms (4 allocs: 9.945 KiB, 99.92% gc time)
```

quinnj:
[Reviewed by AI]

Scoped review of `f25c33a`: no blocker found in the new writer paths. On Julia 1.12.6, 430 extra checks passed for byte/IO/file equivalence, Unicode, large strings, custom lowering, JSON Lines, invalid options, and replacing an existing longer file (including integer payloads). The writer suite also passed all 286 checks after loading its shared fixtures.

The full suite is not green with current dependencies: two existing parser assertions expect positive zero for `-0e291` / `-0e292`, while Parsers 2.8.8 returns negative zero. I reproduced those results on unchanged base `ac46743`, so this is not introduced here. The proposed 0.21 backport was not tested.

PR #384, opened 2025-10-04
Vendor Base.Ryu

nhz2:
From what I can tell, `Base.Ryu` is internal, so it may be removed in a future Julia version.

This PR vendors that code and other internal functions from Base that Ryu uses.


quinnj:
> The point of vendoring this code is that `Base.Ryu` is free to evolve without breaking JSON.jl

Base is already free to evolve? I know package eval is run to check breakage on things, but for something internal like Ryu, I don't think it would stop changes from happening.

> There is no need for this to be kept in sync with Base because it is its own self-contained separate copy.

That sounds problematic to me; bug fixes or even worse, security issues, would then not flow through to JSON eventually via upgrades.

> A check on the existence of `Base.Ryu` doesn't help, because the functions may still exist, but what they do may change.

It seems practical to add sophisticated checks via reflection on the exact methods + shape expected and use if present, and if not exact matching, fall back to non-vendored. It's notable since I wrote the original Ryu implementation, that the main interfaces for the used functions hasn't changed (> 6 years ago).

> JSON.jl is also used to develop Julia itself, so JSON.jl cannot react to changes in Base internals like a normal package.

That doesn't seem particularly fair or relevant here; if Base is using JSON.jl to develop Julia, then _they_ need to be ready to account for changes/breakage in the package like if they relied on any other package. As far as I'm aware, there's no formal documentation or agreement on how this works or is expected to work, so, in my mind, doesn't exist. If JSON.jl is actually critical in a key way to Base, then it should be promoted to a proper stdlib, which I believe is the more formal way of saying "this is functionality we want to be tied/tracked/evaluated against Base more closely than other packages".

I've opened https://github.com/JuliaLang/julia/issues/59763 to further discuss formally exposing an API to the Ryu internal functions and I think that will provide a practical answer here.

hhaensel:
I think, @quinnj 's view is right here. JSON is a package maintained by Julia, if Base.Ryu happens to break JSON in the future it can be fixed at that time.

PR #389, opened 2025-10-06
Fix 32 bit support

### LINKED ISSUE #386: JSON@1 does not support 32-bit machines
See test failure: https://github.com/JuliaIO/JSON.jl/pull/385

Found when trying to update a dependent package: https://github.com/fredo-dedup/JSONSchema.jl/pull/67.

Given the fairly central nature of JSON.jl to the Julia ecosystem, dropping x86 support (if this was intentional) is a fairly wide reaching decision.

nhz2:
Fixes #386

### DIFF
--- a/src/lazy.jl
+++ b/src/lazy.jl
@@ -204,7 +204,7 @@ StructUtils.nulllike(::StructUtils.StructStyle, x::LazyValues) = gettype(x) == J
 
 # core method that detects what JSON value is at the current position
 # and immediately returns an appropriate LazyValue instance
-function _lazy(buf, pos, len, b, opts, isroot=false)
+function _lazy(buf, pos::Int, len, b, opts, isroot=false)
     if opts.jsonlines
         return LazyValue(buf, pos, JSONTypes.ARRAY, opts, isroot)
     elseif b == UInt8('{')
@@ -256,7 +256,7 @@ end
 # returns a `pos` value that notes the next position where parsing should continue
 # this is essentially the `StructUtils.applyeach` implementation for LazyValues w/ type OBJECT
 function applyobject(keyvalfunc, x::LazyValues)
-    pos = getpos(x)
+    pos::Int = getpos(x)
     buf = getbuf(x)
     len = getlength(buf)
     opts = getopts(x)
@@ -356,7 +356,7 @@ end
 # returns a `pos` value that notes the next position where parsing should continue
 # this is essentially the `StructUtils.applyeach` implementation for LazyValues w/ type ARRAY
 function applyarray(keyvalfunc, x::LazyValues)
-    pos = getpos(x)
+    pos::Int = getpos(x)
     buf = getbuf(x)
     len = getlength(buf)
     opts = getopts(x)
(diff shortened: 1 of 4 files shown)


src/lazy.jl
```diff
@@ -256,7 +256,7 @@ end
 # returns a `pos` value that notes the next position where parsing should continue
 # this is essentially the `StructUtils.applyeach` implementation for LazyValues w/ type OBJECT
 function applyobject(keyvalfunc, x::LazyValues)
-    pos = getpos(x)
+    pos::Int = getpos(x)
```
odow, line 259: These shouldn't be necessary, because the fields are all `::Int`:

https://github.com/JuliaIO/JSON.jl/blob/3f7d15f6d6acb54a1abfa10abd8fcad9509be908/src/lazy.jl#L134

src/lazy.jl
```diff
@@ -627,14 +627,14 @@ isbigfloat(x::NumberResult) = x.tag == BIGFLOAT
             # if we overflowed, then let's try BigFloat
             bres = Parsers.xparse2(BigFloat, buf, startpos, len)
             if !Parsers.invalid(bres.code)
-                return NumberResult(bres.val), startpos + bres.tlen
+                return NumberResult(bres.val), Int(startpos + bres.tlen)
```
odow, line 630:
```suggestion
                return NumberResult(bres.val), startpos + Int(bres.tlen)
```

Ah, this is the problem. `bees.tlen` is an `::Int64`:

https://github.com/JuliaData/Parsers.jl/blob/6352c4bbce6f9df8cea268a39a3c43e1ee4944e7/src/Parsers.jl#L31

src/lazy.jl
nhz2: Yes, I found this annotation helpful for documenting the type of `pos` and making sure it doesn't change.

src/lazy.jl
```diff
@@ -627,14 +627,14 @@ isbigfloat(x::NumberResult) = x.tag == BIGFLOAT
...
-                return NumberResult(bres.val), startpos + bres.tlen
+                return NumberResult(bres.val), startpos + Int(bres.tlen)
             end
         end
         if Parsers.invalid(res.code)
             error = InvalidNumber
             @goto invalid
         end
-        return NumberResult(res.val), startpos + res.tlen
+        return NumberResult(res.val), Int(startpos + res.tlen)
```
odow, line 637: Somewhat tangential: but what if this overflows to a negative position? I guess this means that on 32-bit systems we can't read files larger than 2 GB?

src/lazy.jl
nhz2: Yes, I am ok with converting (it might catch some overflow errors as a side benefit).
The convert will happen on all assignments to `pos`, including `pos = (ret isa Int && ret > pos) ? ret : skip(val)` or the assignments to `pos` inside the `@nextbyte` macro.

src/lazy.jl
odow: Doing the `Int(startpos + res.tlen)` will throw an InexactError. Maybe ignore for now. Someone reading a massive file on a 32-bit system is asking for trouble anyway.

quinnj (approved):
Thanks @nhz2, this looks great! Happy to merge after removing the couple of argument type annotations.

PR #391, opened 2025-10-06
Audit use of GC.@preserve

jakobnissen:
Fix one use of GC.@preserve and add TODOs where an invalid pointer escapes GC.@preserve.

@quinnj : Retaining a pointer beyond a `GC.@preserve` is undefined behaviour in Julia, and I've seen at least one case where it causes a real miscompilation.
I've opted not to fix the use of `parsestring` in this PR since that's a larger refactoring that you're best equipped to do. But I've tried to find all the places in the codebase where it's used.

### DIFF
--- a/src/lazy.jl
+++ b/src/lazy.jl
@@ -270,6 +270,7 @@ function applyobject(keyvalfunc, x::LazyValues)
     b == UInt8('}') && return pos + 1
     while true
         # parsestring returns key as a PtrString
+        # TODO: Key contains an invalid pointer here
         key, pos = @inline parsestring(LazyValue(buf, pos, JSONTypes.STRING, opts, false))
         @nextbyte
         if b != UInt8(':')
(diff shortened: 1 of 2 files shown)


src/lazy.jl
```diff
@@ -440,7 +441,18 @@ function Base.convert(::Type{T}, x::PtrString) where {T <: Enum}
     throw(ArgumentError("invalid `$T` string value: \"$sym\""))
 end
 
-Base.:(==)(x::PtrString, y::AbstractString) = x.len == sizeof(y) && ccall(:memcmp, Cint, (Ptr{UInt8}, Ptr{UInt8}, Csize_t), x.ptr, pointer(y), x.len) == 0
+function Base.:(==)(x::PtrString, y::AbstractString)
+    # If the string is backed by UInt8 codeunits, the `sizeof` computation below is correct
+    return if codeunit(y) == UInt8
+        x.len == sizeof(y) || return false
+        ref = Base.cconvert(Ptr{UInt8}, y)
+        cmp = GC.@preserve ref ccall(:memcmp, Cint, (Ptr{UInt8}, Ptr{UInt8}, Csize_t), x.ptr, Base.unsafe_convert(Ptr{UInt8}, ref), x.len)
```
yuyichao, line 449: The manual preserve here seems to be exactly equivalent to simply passing y to ccall directly.

yuyichao:
> It's my understanding that function arguments are GC-preserved/rooted

no this is not the case. (And especially so due to inlining)

yuyichao:
> In perhaps simpler terms, very early in our parsing entrypoints (JSON.lazy, JSON.parse, etc) we create a JSON.LazyValue with an immutable buf field, which is guaranteed to be valid until the end of the entrypoint call. This ensuring that buf is never modified and always live == PtrString usage is safe, as long as its usage is indeed always temporary to its scope (we convert it earlier than we necessarily need to in some places to ensure it never "escapes").

otoh, assuming these are true, is is perfectly valid to ensure the liveness of the object you’ve taken pointer from at the top level and rely on that for all the lower level users, if it is more convenient to do so. However, the only way to guarantee liveness locally outside of a ccall is by using gc preserve.

PR #392, opened 2025-10-06
Audit use of `ccall`

jakobnissen:
The marked used of `ccall` uses a private C function in code that is supposed to be forward compatible. This in private API that may be removed, or change in future versions of Julia

I suggest making issues at the Julia repo to request new features to unsafely create a `String` from a `Memory`.

### DIFF
--- a/src/lazy.jl
+++ b/src/lazy.jl
@@ -418,6 +418,7 @@ if VERSION < v"1.11"
     _tostr(m::Vector{UInt8}, slen) = ccall(:jl_array_to_string, Ref{String}, (Any,), resize!(m, slen))
 else
     mem(n) = Memory{UInt8}(undef, n)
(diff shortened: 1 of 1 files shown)


src/lazy.jl
```diff
@@ -430,6 +431,7 @@ function Base.convert(::Type{String}, x::PtrString)
     return unsafe_string(x.ptr, x.len)
 end
 
+# TODO: This ccall uses a private C function.
 Base.convert(::Type{Symbol}, x::PtrString) = ccall(:jl_symbol_n, Ref{Symbol}, (Ptr{UInt8}, Int), x.ptr, x.len)
```
KristofferC, line 435: Why is this even a convert method? You cannot convert normal strings to Symbols...

PR #396, opened 2025-10-11
Add support for parsing numbers as Int64 to JSONX

JamesWrigley:
IJulia recently added a Python integration that involves passing parsed JSON to various Python libraries (https://github.com/JuliaLang/IJulia.jl/pull/1190). Sometimes they can be picky about types, so this PR adds support for parsing as Int64 where possible. Also popped in a `@view` in `parse_number()` to reduce allocations.

Unfortunately I'd say this is a breaking change. Don't think it makes sense to modify the JSON.jl version number, but we could keep a mini-changelog in the README?

Created with help from Claude :robot:

### DIFF
--- a/vendor/README.md
+++ b/vendor/README.md
@@ -19,7 +19,7 @@ A simple, no-dependency JSON parser that can be vendored (copied/pasted) into ot
 **Reading (JSON → Julia):**
 - `null` → `nothing`
 - `true`/`false` → `Bool`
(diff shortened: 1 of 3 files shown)


quinnj (approved):
Sorry I forgot about this. This LGTM. While breaking, I think it's relatively minor and as you're the only known user of JSONX so far, we'll give you preference here ;)

I agree a changelog in this directory would probably be helpful.

PR #408, opened 2025-11-14
Add new Null/Omit sentinels for overriding omit_null behavior per-field

quinnj:


aviatesk (commented):
Thanks for supporting this!

PR #420, opened 2025-12-18
Fix getindex(::Object{String}, ::Symbol) and vice versa

hhaensel:
Currently `haskey(::Object)` is true for both String and Symbol, but `getindex()` doesn't get it.

```julia-repl
julia> o = JSON.parse("""{"a": "a"}""")
julia> haskey(o, :a)
true
julia> o[:a]
ERROR: MethodError: no method matching find_node_by_key(::JSON.Object{String, Any}, ::Symbol)
The function `find_node_by_key` exists, but no method is defined for this combination of argument types.
```
This PR fixes that issue.


quinnj (approved):
Thanks!

PR #427, opened 2026-01-22
add JSONText methods for Symbol and JSONText

hhaensel:
In some contexts it would be beneficial to have JSONText constructors for Symbol and JSONText arguments. What do you think about the following defintions?
```julia
JSON.JSONText(js::Symbol) = JSONText(String(js))
JSON.JSONText(js::JSONText) = js
```
I think it doesn't hurt and only adds methods that a user would expect to work as defined.


src/JSON.jl
```diff
@@ -70,6 +70,8 @@ x = JSON.parse("[1,2,3]", JSONText)
 struct JSONText
     value::String
 end
+JSON.JSONText(js::Symbol) = JSONText(String(js))
```
quinnj: This definition feels really weird/unnecessary to me (who has valid json as a symbol?). I think I'd prefer to not support for now. I think the other definition is fine.

hhaensel: I've written a Julia to JS expression conversion where I use Symbols to mean js variables, e.g.
```julia-repl
julia> @jsexpr(:x^2 + :b * :y + 2)
JSExpr("(((x ** 2) + (b * y)) + 2)")
```
In this context it comes in very handy to automatically convert Symbols to JSONText.

quinnj:
[Reviewed by AI]

One blocker at `c39cb680`: the `i` flag cannot handle quoted JSON keys or string values. On Julia 1.12.6, `JSON.js"""{"a":$(1 + 2)}"""i` fails during macro expansion. Wrapping the raw text in `Meta.parse("\"$s\"")` produces `"{"a":$(1 + 2)}"`, so the inner quotes terminate the Julia string.

Please parse interpolation without treating the JSON quotes as Julia delimiters, and add macro tests for quoted keys, quoted values, backslashes, and interpolation. The current new test covers only `JSONText(::JSONText)`.

hhaensel:
Oh, I didn't realize that I didn't fix this in the PR. I added triple quotes for parsing, I think that should do.
And I added respective tests.

src/JSON.jl
```diff
@@ -111,6 +111,35 @@ x = JSON.parse("[1,2,3]", JSONText)
...
+```
+"""
+macro js_str(s)
+    :( JSONText($(esc(s))) )
+end
+
+macro js_str(s, flags)
+    flags == "i" || @warn "Only 'i' flag currently supported (string interpolation)."
+    if 'i' in flags
+        :( JSONText($(esc(Meta.parse("\"\"\"$s\"\"\"")))) )
```
quinnj: [P2] Preserve a closing JSON quote in interpolated root strings

[Reviewed by AI] `JSON.js"\"$(1 + 2)\""i` should produce raw JSON `"3"`, but it throws `ParseError` during expansion: the closing JSON quote meets the synthesized triple-quote delimiter. The same input with trailing whitespace succeeds. Reproduced on Julia 1.10.12 and 1.12.6; the previously reported object example now passes. Please handle this delimiter boundary and add a root-string regression.

quinnj (commented):
[Reviewed by AI] Reviewed this exact head. Focused constructor, macro, read/write, and custom-hook checks passed on Julia 1.10 and 1.12; one remaining root-string interpolation failure below.

PR #445, opened 2026-03-27
[DOC] fix json doc

ogauthe:
Hi! I found that when displaying the doc for `JSON.json` in the console, a `\n` char was not correctly escaped.

This PR fixes the display issue in the doc.


quinnj (approved):
Thanks!

PR #461, opened 2026-05-20
Mark user-facing methods `public`

timholy:
`public` was added in Julia 1.11 and the ExplicitImports package allows one to check whether a package only accesses public names in its dependencies. I've been finding this a useful addition to many of my packages. The help string now shows
```
help?> JSON.parsefile
  │ Warning
  │
  │  The following bindings may be internal; they may change or be removed in future versions:
  │
  │    •  JSON.parsefile

  See parse.
```
This should silence the warning and allow ExplicitImports tests in other packages that use JSON to pass.


quinnj (approved):
Thanks!

PR #470, opened 2026-06-24
vendor/jsonx: improve correctness and performance

fredrikekre:
Some minor things that Claude commented on while I was using the jsonx vendored
implementation. See commit message for details.


quinnj (approved):
Thanks!

PR #479, opened 2026-08-20
FIX InexactError when using allownan with typemax(Int)

### LINKED ISSUE #478: Inexact Error when parsing using typemax(Int) and allownan=true
Hello, 

When I am using JSON.parsefile with allownan=true, I have InexactError when trying to parse an field with typemax(Int). I encountered this bug when trying to load a struct with a field being a TimerOutputs.TimerOutput, that is using typemax for their time_min and time_max fields. 

I wrote below a minimum working example, and paste the associate stack trace. It look like it's loaded as a Float64, and then converted to Int64, but it's not working (see example below on multiple conversion). What it's surprising is that I don't have problem when I don't use allownan=true. 

```
julia> Int64(Float64(typemax(Int64)))
ERROR: InexactError: Int64(9.223372036854776e18)
Stacktrace:
 [1] Int64(x::Float64)
   @ Base ./float.jl:923
 [2] top-level scope
   @ REPL[21]:1
```

Version: JSON v1.7.1, Julia 1.12.6 on MacBook Pro M5 with MacOS 26.5.2

**MWE**
```julia

using JSON
struct Test
    a::Int64
end

t = Test(typemax(Int64))
JSON.json("typemax.json", t;allownan=true )
t2 = JSON.parsefile("typemax.json", Test;allownan=true)
```

**Error:**
```julia

ERROR: InexactError: Int64(9.223372036854776e18)
Stacktrace:
  [1] Int64
    @ ./float.jl:923 [inlined]
  [2] convert
    @ ./number.jl:7 [inlined]
  [3] lift
    @ ~/.julia/packages/StructUtils/019BB/src/StructUtils.jl:421 [inlined]
  [4] lift
    @ ~/.julia/packages/StructUtils/019BB/src/StructUtils.jl:578 [inlined]
  [5] lift
    @ ~/.julia/packages/StructUtils/019BB/src/StructUtils.jl:597 [inlined]
  [6] lift
    @ ~/.julia/packages/JSON/5zJAn/src/parse.jl:439 [inlined]
  [7] lift(style::JSON.JSONReadStyle{JSON.Object{String, Any}, Nothing, StructUtils.DefaultStyle}, ::Type{Int64}, x::JSON.LazyValue{Vector{UInt8}}, tags::@NamedTuple{})
    @ JSON ~/.julia/packages/JSON/5zJAn/src/parse.jl:481
  [8] _make
    @ ~/.julia/packages/StructUtils/019BB/src/StructUtils.jl:1065 [inlined]
  [9] make
    @ ~/.julia/packages/StructUtils/019BB/src/StructUtils.jl:1008 [inlined]
 [10] macro expansion
    @ ~/.julia/packages/StructUtils/019BB/src/StructUtils.jl:1430 [inlined]
 [11] applyfield!
    @ ~/.julia/packages/StructUtils/019BB/src/StructUtils.jl:1424 [inlined]
 [12] FieldSink
    @ ~/.julia/packages/StructUtils/019BB/src/StructUtils.jl:1450 [inlined]
 [13] applyobject(keyvalfunc::StructUtils.FieldSink{Test, JSON.JSONReadStyle{JSON.Object{String, Any}, Nothing, StructUtils.DefaultStyle}, Memory{Any}, StructUtils.NoFieldMetadata, Base.RefValue{Int64}}, x::JSON.LazyValue{Vector{UInt8}})
    @ JSON ~/.julia/packages/JSON/5zJAn/src/lazy.jl:304
 [14] applyeach
    @ ~/.julia/packages/JSON/5zJAn/src/lazy.jl:188 [inlined]
 [15] fillfields!
    @ ~/.julia/packages/StructUtils/019BB/src/StructUtils.jl:1460 [inlined]
 [16] makestruct(style::JSON.JSONReadStyle{JSON.Object{String, Any}, Nothing, StructUtils.DefaultStyle}, ::Type{Test}, source::JSON.LazyValue{Vector{UInt8}})
    @ StructUtils ~/.julia/packages/StructUtils/019BB/src/StructUtils.jl:1493
 [17] make
    @ ~/.julia/packages/StructUtils/019BB/src/StructUtils.jl:1138 [inlined]
 [18] _parse
    @ ~/.julia/packages/JSON/5zJAn/src/parse.jl:273 [inlined]
 [19] parse(buf::Vector{UInt8}, ::Type{Test}; dicttype::Type{JSON.Object{String, Any}}, null::Nothing, style::StructUtils.DefaultStyle, unknown_fields::Symbol, kw::@Kwargs{jsonlines::Bool, allownan::Bool})
    @ JSON ~/.julia/packages/JSON/5zJAn/src/parse.jl:257
 [20] parse
    @ ~/.julia/packages/JSON/5zJAn/src/parse.jl:257 [inlined]
 [21] parse
    @ ~/.julia/packages/JSON/5zJAn/src/parse.jl:245 [inlined]
 [22] #45
    @ ~/.julia/packages/JSON/5zJAn/src/parse.jl:242 [inlined]
 [23] open(f::JSON.var"#45#46"{Test, Nothing, @Kwargs{allownan::Bool}, String}, args::String; kwargs::@Kwargs{})
    @ Base ./io.jl:412
 [24] open
    @ ./io.jl:409 [inlined]
 [25] #parsefile#43
    @ ~/.julia/packages/JSON/5zJAn/src/parse.jl:242 [inlined]
 [26] top-level scope
    @ REPL[12]:1
```

Kiwy3:
## Summary
- correct parsenumber function and check_special macro to fix the bug
- add 2 tests (for Int64 and Int128) to ensure future versions will work with this

## Root Cause
When using allownan, any number was parse as Float, and then convert as Int if needed. But the parsing as Float64 was rounded from 2e63 -1 to 2e63. 

Notice: it's my first PR, so I may have done some mistakes. I ran all tests locally, and it was working. 

Fixes #478

### DIFF
--- a/src/lazy.jl
+++ b/src/lazy.jl
@@ -539,16 +539,16 @@ macro check_special(special, value)
         b = getbyte(buf, pos)
         bytes = codeunits($special)
         i = 1
-        while b == @inbounds(bytes[i])
+        while i <= length(bytes) && b == @inbounds(bytes[i])
             pos += 1
             i += 1
-            i > length(bytes) && break
-            if pos > len
-                error = UnexpectedEOF
-                @goto invalid
+            if i <= length(bytes)
+                if pos > len
+                    error = UnexpectedEOF
+                    @goto invalid
+                end
+                b = getbyte(buf, pos)
             end
-            b = getbyte(buf, pos)
-            i += 1
         end
         if i > length(bytes)
             return NumberResult($value), pos
--- a/src/lazy.jl
+++ b/src/lazy.jl
@@ -589,84 +589,103 @@ isbigfloat(x::NumberResult) = x.tag == BIGFLOAT
     isneg = isfloat = overflow = false
 
     if opts.allownan
-        @check_special(opts.nan, NaN)
-        @check_special(opts.inf, Inf)
-        @check_special(opts.ninf, -Inf)
-        # reset after any failed partial match (see note above)
-        pos = startpos
-        b = getbyte(buf, pos)
-    end
-
-    val = Int64(0)
-    isneg = b == UInt8('-')
-    if isneg || b == UInt8('+') # spec doesn't allow leading +, but we do
-        pos += 1
-        if pos > len
-            error = UnexpectedEOF
-            @goto invalid
-        end
-        b = getbyte(buf, pos)
-    end
-    # Parse integer part, check for leading zeros (invalid JSON)
-    if b == UInt8('0')
-        pos += 1
-        if pos <= len
-            b = getbyte(buf, pos)
-            if UInt8('0') <= b <= UInt8('9')
-                error = InvalidNumber
+        # If this doesn't start like an integer, let Parsers handle
+        # NaN/Inf/-Inf and the configured special values.
+        if b == UInt8('-') || b == UInt8('+')
+            if pos == len
+                error = UnexpectedEOF
                 @goto invalid
             end
+            b = getbyte(buf, pos + 1)
+            isfloat = !(UInt8('0') <= b <= UInt8('9'))
+        else
+            isfloat = !(UInt8('0') <= b <= UInt8('9'))
         end
-    elseif UInt8('1') <= b <= UInt8('9')
-        while UInt8('0') <= b <= UInt8('9')
-            digit = Int64(b - UInt8('0'))
-            if val > INT64_OVERFLOW_VAL || (val == INT64_OVERFLOW_VAL && digit > INT64_OVERFLOW_DIGIT)
-                overflow = true
-                break
-            end
-            val = Int64(10) * val + digit
-            pos += 1
-            pos > len && break
-            b = getbyte(buf, pos)
-        end
-        if overflow
-            bval = BigInt(val)
-            while UInt8('0') <= b <= UInt8('9')
-                digit = BigInt(b - UInt8('0'))
-                bval = BigInt(10) * bval + digit
-                pos += 1
-                pos > len && break
-                b = getbyte(buf, pos)
-            end
-        end
-    elseif opts.allownan
-        # not a digit/sign start and no special token matched above -
-        # let Parsers' own float lexer take a shot (handles any
-        # remaining native spellings it knows about)
-        isfloat = true
-    else
-        error = InvalidNumber
-        @goto invalid
     end
-    # Check for decimal or exponent
-    if !isfloat && (b == UInt8('.') || b == UInt8('e') || b == UInt8('E'))
-        isfloat = true
-        if b == UInt8('.')
+
+    if !isfloat
+        val = Int64(0)
+        b = getbyte(buf, pos)
+        isneg = b == UInt8('-')
+
+        if isneg || b == UInt8('+') # spec doesn't allow leading +, but we do
             pos += 1
             if pos > len
                 error = UnexpectedEOF
                 @goto invalid
             end
             b = getbyte(buf, pos)
-            if !(UInt8('0') <= b <= UInt8('9'))
-                error = InvalidNumber
-                @goto invalid
+        end
+
+        # Parse integer part, check for leading zeros (invalid JSON)
+        if b == UInt8('0')
+            pos += 1
+            if pos <= len
+                b = getbyte(buf, pos)
+                if UInt8('0') <= b <= UInt8('9')
+                    error = InvalidNumber
+                    @goto invalid
+                end
+            end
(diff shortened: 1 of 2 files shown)


quinnj:
Thanks for the clear writeup and for jumping on a first PR the same day.

## Diagnosis

With `allownan=true`, `parsenumber` currently skips the Int64/BigInt lexer and always goes through `Parsers.xparse2(Float64, ...)`. `Float64(typemax(Int64))` rounds to `2^63`, and converting that to `Int64` throws `InexactError`. Same class of bug for any integer with `|n| > 2^53` (silent wrong value) and for `typemax(Int128)` / `typemax(UInt64)`.

Always running the integer lexer, and only taking the float path for actual floats / specials, is the right fix and the right layer.

## `@check_special` rewrite

This part I don't want to take as-is.

Master already double-increments `i`, so multi-char tokens almost never match — `NaN` / `Inf` / `Infinity` actually work via the Parsers fallback. The new loop matches correctly, but a prefix of the configured token at EOF still `@goto invalid`. Default `inf` is `"Infinity"`, so:

```julia
JSON.parse("Inf"; allownan=true)    # UnexpectedEOF after this PR
JSON.parse("[Inf]"; allownan=true)  # still works (`]` → failed match → Parsers)
```

That's an inconsistency, and CI doesn't see it because the existing test overrides `inf="Inf"`. Docs advertise `allownan` as allowing `NaN`, `Inf`, `-Inf`.

Also a behavior change worth calling out: untyped `JSON.parse("123"; allownan=true)` (and `+1`) will now return `Int64` instead of `Float64`. That's the correct consistency with `allownan=false`.

## Suggested fix

Either of these is fine:

1. **Smaller patch:** don't touch `@check_special`. Remove the `if !opts.allownan` wrapper, add `elseif opts.allownan; isfloat = true`, and change `if isfloat || opts.allownan` to `if isfloat`. Specials stay on the Parsers path like today.

2. **Keep the new structure,** but treat mid-token EOF as a failed match (then reset, which you already do) so `"Inf"` can fall through to Parsers as your comment intends.

## Tests to add

- the original struct-field MWE from #478
- default `"NaN"` / `"Infinity"` / `"-Infinity"` *and* `"Inf"` / `"-Inf"` without overriding `inf`/`ninf`
- nested: `JSON.parse("[Inf,NaN,-Infinity]"; allownan=true)`
- `typemin(Int64)`, `2^53+1`, maybe `typemax(UInt64)`
- untyped `JSON.parse(string(typemax(Int64)); allownan=true) === typemax(Int64)`

Minor: trailing spaces on the new test lines, and "see note above" has no note.

Happy to merge once the Inf prefix case is fixed and those tests are in — this is a solid first PR.

[analysis & commentary by AI; reviewed by quinnj]

Kiwy3:
Thanks for your fast feedback!

I reworked my PR without modifying the `check_special` macro, and added the mentioned tests, along with two additional tests to make sure the correct errors are thrown when using `allownan`.

I tried a few different approaches to handle all the specific cases covered by the tests, and I believe the current version should work correctly.

I also removed all the trailing spaces I had added previously, and I believe all the comments should now be clear and up to date.

If everything looks good to you, I'm happy with the current version. Otherwise, I'm happy to make any further changes if needed.

src/lazy.jl
```diff
@@ -587,8 +587,23 @@ isbigfloat(x::NumberResult) = x.tag == BIGFLOAT
...
+                error = UnexpectedEOF
+                @goto invalid
+            end
+            b = getbyte(buf, pos + 1)
+            isfloat = !(UInt8('0') <= b <= UInt8('9'))
+        else
+            isfloat = !(UInt8('0') <= b <= UInt8('9'))
+        end
+    end
+    if !isfloat
```
quinnj: [Reviewed by AI] P2: Keep configured numeric special tokens ahead of integer parsing. For example, `JSON.parse("0"; allownan=true, nan="0")` and `JSON.parse("1"; allownan=true, inf="1")` returned NaN/Inf on the original base; this branch returns integers 0/1 because the new integer path bypasses `@check_special`. Both baseline controls pass with matched dependencies. Current master already preserves these overrides through #480, so this logic should not replace that implementation.

quinnj (commented):
[Reviewed by AI] Rechecked head 527057f. The original typed-integer failure is now covered by merged #480; current master 3e4cb97 passes the original struct case and Int64/UInt64/Int128 limit probes. This branch conflicts with current master in src/lazy.jl and test/parse.jl. Its full local suite also hits two signed-zero expectations that fail on the original base with the same dependencies. The new regression below is independent of those existing failures.

